PR #3, opened 2024-11-17
Small fixes all around

nmetulev:
* deleting a model from the cache outside of the app now also removes it from the app
* ensure cache folder is always created
* fixed models not showing up in search
* changed default cache name to "aigallery"
* fixed view model card button not working in download list
* updated version to 0.0.1
* Added manage models link in models page


AIDevGallery/App.xaml.cs
```diff
@@ -122,48 +122,33 @@ private void GenerateSearchIndex()
...
-                    {
-                        SearchIndex.Add(new SearchResult() { Label = modelDetails.Name, Description = modelDetails.Description, Tag = modelType });
-                    }
-                    else if (ModelTypeHelpers.ModelFamilyDetails.TryGetValue(modelType, out var modelFamily))
+                    if (innerItems?.Count > 0)
                     {
-                        SearchIndex.Add(new SearchResult() { Label = modelFamily.Name, Description = modelFamily.Description, Tag = modelType });
+                        foreach (var childNavigationItem in innerItems)
+                        {
+                            if (ModelTypeHelpers.ModelGroupDetails.TryGetValue(childNavigationItem, out var modelGroup))
```
azchohfi, line 139: So now we can't search for specific models, only for families? Was that the intent?

nmetulev: Yes, the view is for the model family anyway - before we couldn't search for most models (phi was returning no results). A user would be looking for "phi 3" not "phi 3 CPU"

PR #23, opened 2024-11-20
Update README.md

AndrewKeepCoding:
Added missing emojis on sub-headers


AndrewKeepCoding:
Just wanted to point out that there are missing emojis and there's no need to use the emojis I added.
Please updated them if necessary. 🙂

nmetulev (approved):

PR #26, opened 2024-11-21
Adding some hero imagery

niels9001:


azchohfi (approved):

docs/images/HeroImage0.png
azchohfi, line 1: Most of these files are not referenced in any Markdown. Is that correct?

niels9001: Yup

PR #41, opened 2024-11-27
Added QNN EP for ARM64 builds.

azchohfi:


azchohfi:
Not ready for review since the exported projects are not great yet... I'm not sure what to do there...

AIDevGallery/Models/ModelCompatibility.cs
```diff
@@ -23,7 +23,8 @@ public static ModelCompatibility GetModelCompatibility(ModelDetails modelDetails
 
         ModelCompatibilityState compatibility;
         if (modelDetails.HardwareAccelerators.Contains(HardwareAccelerator.CPU) ||
-            (modelDetails.HardwareAccelerators.Contains(HardwareAccelerator.DML) && modelDetails.SupportedOnQualcomm == true))
+            (modelDetails.HardwareAccelerators.Contains(HardwareAccelerator.DML) && modelDetails.SupportedOnQualcomm == true) ||
```
nmetulev: We should disable all DML models on Arm64

AIDevGallery/Samples/SharedCode/EmbeddingGenerator.cs
```diff
@@ -44,6 +45,15 @@ public EmbeddingGenerator(string modelPath, HardwareAccelerator hardwareAccelera
         {
             _sessionOptions.AppendExecutionProvider_DML(DeviceUtils.GetBestDeviceId());
         }
+        else if (hardwareAccelerator == HardwareAccelerator.QNN)
+        {
+            Dictionary<string, string> options = new()
+            {
+                { "backend_path", "QnnHtp.dll" }
```
nmetulev: we should include the high performance kv here too

AIDevGallery/Samples/ModelsDefinitions/embeddings.modelgroup.json
```diff
@@ -18,7 +18,8 @@
             "Description": "This is an embedding model executed on the CPU or GPU",
             "HardwareAccelerator": [
               "CPU",
-              "DML"
+              "DML",
+              "QNN"
```
nmetulev: we need to do some testing here before enabling the models for qnn - I'm seeing both the cpu and npu running which might mean the model is falling back to cpu and it's not optimized for qnn

AIDevGallery/Samples/SharedCode/EmbeddingGenerator.cs
nmetulev: Added

AIDevGallery/Models/ModelCompatibility.cs
nmetulev: disabled

AIDevGallery/Samples/ModelsDefinitions/embeddings.modelgroup.json
nmetulev: removed qnn from embeddings models

nmetulev:
@azchohfi, I made few changes to address my initial comments, please take a look.

In addition, the exported projects still need work to include ORT.QNN instead of ORT.DML. Also, the execution provider in the exported project defaults to CPU even if NPU is selected in the app. Once this is added, I would be good to merge this in

AIDevGallery/Samples/SharedCode/EmbeddingGenerator.cs
azchohfi: You removed QNN from the embedding models, so this if statement should be removed.

nmetulev: I kept it in there for when we do add QNN - doesn't break anything

AIDevGallery/Samples/SharedCode/EmbeddingGenerator.cs
```diff
@@ -44,6 +45,17 @@ public EmbeddingGenerator(string modelPath, HardwareAccelerator hardwareAccelera
...
+        {
+            Dictionary<string, string> options = new()
+            {
+                { "backend_path", "QnnHtp.dll" },
+                { "htp_performance_mode", "high_performance" },
+                { "htp_graph_finalization_optimization_mode", "3" }
+            };
+            _sessionOptions.AppendExecutionProvider("QNN", options);
+            _chunkSize = 8;
+        }
```
zateutsch, line 58: Should be removed if we're not doing NPU for all-Mini

zateutsch (approved):
Approved with one minor comment.

PR #43, opened 2024-12-04
Fix crash when running and exiting the generated project

brunosonnino:
Fix crash when running and exiting the generated project
  Added OgaHandle in the generated MainWindow
Minor fixes in the Custom System Prompt sample
Bump OnnxRuntimeGenAI version to 0.5.2

### DIFF
--- a/AIDevGallery/Samples/Open Source Models/Language Models/CustomSystemPrompt.xaml.cs
+++ b/AIDevGallery/Samples/Open Source Models/Language Models/CustomSystemPrompt.xaml.cs
@@ -43,7 +43,7 @@ internal sealed partial class CustomSystemPrompt : BaseSamplePage
     protected override async Task LoadModelAsync(SampleNavigationParameters sampleParams)
     {
         model = await sampleParams.GetIChatClientAsync();
-        InputTextBox.MaxLength = GenAIModel.DefaultMaxLength;
+        InputTextBox.MaxLength = chatOptions.MaxOutputTokens ?? 0;
         sampleParams.NotifyCompletion();
     }
 
@@ -180,18 +180,19 @@ internal sealed partial class CustomSystemPrompt : BaseSamplePage
 
     private void InputBox_Changed(object sender, TextChangedEventArgs e)
     {
-        var maxInputLength = GenAIModel.DefaultMaxLength;
-
         var inputLength = InputTextBox.Text.Length;
         if (inputLength > 0)
         {
-            InputTextBox.Description = $"{inputLength} of {maxInputLength}";
-            if (inputLength >= maxInputLength)
+            if (inputLength >= chatOptions.MaxOutputTokens)
             {
-                InputTextBox.Description += ". Max characters reached.";
+                InputTextBox.Description = $"{inputLength} of {chatOptions.MaxOutputTokens}. Max characters reached.";
+            }
+            else
+            {
+                InputTextBox.Description = $"{inputLength} of {chatOptions.MaxOutputTokens}";
             }
 
-            GenerateButton.IsEnabled = inputLength <= maxInputLength;
+            GenerateButton.IsEnabled = inputLength <= chatOptions.MaxOutputTokens;
         }
         else
         {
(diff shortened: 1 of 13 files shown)


AIDevGallery/Samples/Open Source Models/Language Models/CustomSystemPrompt.xaml.cs
```diff
@@ -43,7 +43,7 @@ public CustomSystemPrompt()
     protected override async Task LoadModelAsync(SampleNavigationParameters sampleParams)
     {
         model = await sampleParams.GetIChatClientAsync();
-        InputTextBox.MaxLength = chatOptions.MaxOutputTokens ?? 0;
+        InputTextBox.MaxLength = GenAIModel.DefaultMaxLength;
```
azchohfi: This change is unrelated to the fix, right?
What behavior changes are expected? `MaxOutputTokens` is initialized to the same value. We need to make the samples more generic, and this change seems to go against that so we should try to move away from coupling the samples with the GenAIModel constants, if possible.

Directory.Packages.props
```diff
@@ -18,9 +18,9 @@
     <PackageVersion Include="CommunityToolkit.WinUI.Converters" Version="8.1.240916" />
     <PackageVersion Include="CommunityToolkit.WinUI.UI.Controls.Markdown" Version="7.1.2" />
     <PackageVersion Include="Microsoft.Build" Version="17.11.4" />
-    <PackageVersion Include="Microsoft.ML.OnnxRuntime.DirectML" Version="1.20.0" />
+    <PackageVersion Include="Microsoft.ML.OnnxRuntime.DirectML" Version="1.20.1" />
     <PackageVersion Include="Microsoft.ML.OnnxRuntime.Extensions" Version="0.13.0" />
-    <PackageVersion Include="Microsoft.ML.OnnxRuntimeGenAI.DirectML" Version="0.5.0" />
+    <PackageVersion Include="Microsoft.ML.OnnxRuntimeGenAI.DirectML" Version="0.5.2" />
```
azchohfi, line 23: I think we still, have a regression with this version:
https://github.com/microsoft/onnxruntime-genai/issues/1071

azchohfi (dismissed):

AIDevGallery/Samples/Open Source Models/Language Models/CustomSystemPrompt.xaml.cs
brunosonnino: All other language samples use GenAIModel.DefaultMaxLength, and ChatOptions.MaxOutputTokens are the maximum output tokens, not the maximum input length. 
It would make sense (eventually) to add a new slider, but I don't think it's needed

azchohfi: The other samples use the constants since they don't have an instance of `chatOptions`, but I see what you mean.
Lets open a separate issue and deal with this on a separate PR. This is important, but just not part of the same crash/issue.

Directory.Packages.props
brunosonnino: I tested it here, it works for me:
image

AIDevGallery/Samples/SharedCode/GenAIModel.cs
```diff
@@ -84,6 +85,11 @@ private GenAIModel(string modelDir)
         return model;
     }
 
+    public static void InitializeGenAI()
+    {
+        _ogaHandle = new OgaHandle();
```
nmetulev, line 90: Could we not just auto do this on each GenAIModel instance so it auto handles this both for the gallery and for the generated projects? If we can avoid having devs to have to call this, we should.

brunosonnino: No, it doesn't work. If you add this to GenAIModel, ogahandle will be disposed when the model is disposed and next time it will crash. In the gallery it was added to Program.cs.
It should be initialized and disposed only once in the app

nmetulev (approved):

PR #49, opened 2024-12-11
Fixes download for stable diffusion

nmetulev:
Fixes an issue (introduced in #5) where if files are in subfolders, then download will fail.


zateutsch (approved):
Just bumped into this myself, lgtm, thanks

PR #60, opened 2024-12-13
Added filter in scenario view

nmetulev:
Added filter in scenario view to filter to NPU/GPU and eventually WCR APIs

https://github.com/user-attachments/assets/b179978b-c457-4964-bc9a-470d80b5f6ba

### DIFF
--- a/AIDevGallery/App.xaml
+++ b/AIDevGallery/App.xaml
@@ -9,6 +9,7 @@
             <ResourceDictionary.MergedDictionaries>
                 <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
                 <ResourceDictionary Source="/Styles/Button.xaml" />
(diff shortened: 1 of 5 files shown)


niels9001:
@nmetulev We could probably turn this into an `ItemsView` (so we have multi-select) and make the tokens look like the tags that we use in the model selection control. So we'd have 4 buttons that are multi-selectable:

- CPU
- GPU
- NPU
- API

Does that make sense? I'm happy to push the changes to this branch.

nmetulev:
we need to make it clear what multiselect means. does it filter to samples that support all the selected items, or any of the selected items. If you have a good solution to that, then I'm good with this solution

niels9001:
> we need to make it clear what multiselect means. does it filter to samples that support all the selected items, or any of the selected items. If you have a good solution to that, then I'm good with this solution

Yeah.. great point.

I've updated the style for now so we can get this in. It's now a bit more subtle:

image

Issue #13, opened 2024-11-19
[P2 FEATURE] Add circular progress ring when search is loading


gregwoo-microsoft:
Right now, hitting search doesn't show a loading indicator

![Image](https://github.com/user-attachments/assets/0c7224e1-6b98-49f1-ba4a-0567f6b6d208)

PR #72, opened 2025-01-08
[P2 FEATURE] Add circular progress ring when search is loading

Pinguin2001:
Implements: https://github.com/microsoft/ai-dev-gallery/issues/13


niels9001 (approved):
Tested, works well.. Thanks for your contribution @pinguin2001!

PR #77, opened 2025-01-15
#73: Fixed syntax highlighting issue in dark mode for Generate Code sample

### LINKED ISSUE #73: [BUG] Syntax highlighting broken for the generate code sample in dark mode
**Describe the bug**

On the generate code sample, the syntax highlighting seems off in dark mode. When displaying the sample code, syntax highlighting seems to switch fine, but the generated code from the sample does not. Below is an example of generated C# and the sample code side by side in both light & dark mode.

**To Reproduce**
1. Run the generate code sample for C# while your machine is using dark mode
2. Click on the view code button to compare between the generated code and sample code

**Expected behavior**

I would expect the generated code to be highlighted appropriately.

**Screenshots**

Image

Image

**Please complete the following information:**
 - Version 0.0.1.0

zateutsch:
fixes #73 

Fixed code highlighting in dark mode for the generate code sample. I think this will save us an accessibility issue down the road as well.

Moved over code highlighting logic to AppUtils so it could be referenced by the GenerateCode sample and the sample container.

Thanks @michaeljolley for the catch.


AIDevGallery/Samples/Open Source Models/Language Models/GenerateCode.xaml.cs
```diff
@@ -31,16 +32,20 @@ namespace AIDevGallery.Samples.OpenSourceModels.LanguageModels;
     Icon = "\uE8D4")]
 internal sealed partial class GenerateCode : BaseSamplePage
 {
-    private readonly RichTextBlockFormatter formatter = new();
+    private RichTextBlockFormatter formatter;
     private IChatClient? model;
     private CancellationTokenSource? cts;
 
     public ObservableCollection<string> LanguageStrings { get; } = ["C#", "C++", "Java", "Python", "JavaScript", "TypeScript"];
+    private string _currentLanguage = "C#";
```
zateutsch, line 41: Not setting _currentLanguage again

azchohfi:
Ok, tests are fixed, and this PR is actually breaking the build.

AIDevGallery/Utils/AppUtils.cs
```diff
@@ -158,4 +161,32 @@ public static ImageSource GetModelSourceImageFromUrl(string url)
             return new SvgImageSource(new Uri("ms-appx:///Assets/ModelIcons/HuggingFace.svg"));
         }
     }
+
+    public static StyleDictionary GetCodeHighlightingStyleFromElementTheme(ElementTheme theme)
```
azchohfi, line 165: This method is not available to exported projects. That is why the tests are failing. We need another solution here.
Either making this method available for the exported projects, or not calling this method on the exported projects (with `// <exclude>` maybe?).

azchohfi (changes requested):

AIDevGallery/Utils/AppUtils.cs
zateutsch: I'll look into it.

nmetulev:
@niels9001, can I ask you to review?

niels9001:
Tested! Works!

niels9001 (approved):

Issue #12, opened 2024-11-19
[P2 FEATURE] Add error string when search doesn't find any results


gregwoo-microsoft:
![Image](https://github.com/user-attachments/assets/4e34c54e-1f8f-49a0-8cff-9b2579c6499f)

There should be an error message when the search query doesn't yield any results

Pinguin2001:
I can work on this, is a ContentDialog fine, or should a InfoBar be displayed instead?

nmetulev:
Thanks @Pinguin2001. @niels9001, what do you think about how to display this error message?

niels9001:
A TextBlock with message in the results pane would suffice?

Pinguin2001:
Id say either a simple TextBlock or InfoBar

pealmasa:
Not sure if this was understood, but it seems the issue is more "deep" than simply a model not found in search and add a text block or infobar.

See: https://github.com/microsoft/ai-dev-gallery/issues/75

niels9001:
Closed by #116

PR #87, opened 2025-01-16
Fix model search

Rene-Sackers:
I noticed the model search wasn't working, first it crashed when deserializing the JSON because the `trendingScore` of a result was a double (`0.2`) while the model expected an `int`, after that, the search button never finished "loading", because none of the results had a `genai_config.json`.

I fixed the trendingScore issue by changing the property to a double.

I fixed the no results issue by checking if the `ActionBlock` that processes the results has an `InputCount` of 0 after it queues the results for processing, if so, call `actionBlock.Complete();`

I apologize for the large indentation change, I inverted an if statement to reduce nesting in this function, I changed notihing in `AddModelPage.xaml.cs` other than the invert if and addition of
```csharp
        if (actionBlock.InputCount == 0)
        {
            actionBlock.Complete();
        }
```

**Notes:**
Please verify if my call to `TelemetryFactory.Get<ITelemetry>().LogException("HuggingFaceApiSearchFailed_Event", ex);` is correct, I don't fully understand the logging/telemetry in this project.

It would be nice to add a UI hint stating that there are no results, but at least it doesn't crash or infinitely show "searching" now :)

https://github.com/user-attachments/assets/f96f9dbf-9b92-46cc-82b2-82900fef2f68

### DIFF
--- a/AIDevGallery/Models/HFSearchResults.cs
+++ b/AIDevGallery/Models/HFSearchResults.cs
@@ -25,7 +25,7 @@ internal class HFSearchResult
     [JsonPropertyName("likes")]
     public int Likes { get; set; }
     [JsonPropertyName("trendingScore")]
-    public int TrendingScore { get; set; }
+    public double TrendingScore { get; set; }
     [JsonPropertyName("_private")]
     public bool Private { get; set; }
     [JsonPropertyName("sha")]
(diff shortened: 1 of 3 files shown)


Rene-Sackers:
@microsoft-github-policy-service agree

azchohfi:
I agree that we should have a "no results" message. We have an open issue tracking that (#12).
We can still merge this as-is. Thank you very much for your contribution!

Issue #84, opened 2025-01-16
[BUG] Ending Mistral generation puts a <|end|> token in line.


Jaylyn-Barbee:
**Describe the bug**
Title

**To Reproduce**
1. Go to any language sample
2. Select Mistral CPU for the model
3. Allow it to start generating output and before its finished press "Stop"
4. Observe character that gets place in line

**Expected behavior**
It should just end

**Screenshots**
![Image](https://github.com/user-attachments/assets/be318a03-f9f3-4303-a3f3-14f2bac8c5d1)

**Please complete the following information:**
 - App Version: latest version of main for 15:52 on 1/16/25

**Additional context**

PR #89, opened 2025-01-17
[BUGBASH] #84: Fix stop tokens during LM cancellation.

zateutsch:
If cancellation is requested, GenAIModel will now add the correct stop token for each LLM instead of always using <|end|>


Jaylyn-Barbee (approved):
Works and code is clean. LGTM

PR #93, opened 2025-01-18
#51: Copy over sample specific asset files to exported samples

### LINKED ISSUE #51: [BUG] Exported projects don't include relevant assets, like a default image
Image assets aren't being included in project export. Not an urgent error as we are checking the file path, so nothing breaks and they can use the sample with manual image upload.

Noticed this in Segment Streetscapes, but I believe it's true of all samples that have a default image asset:

Image reference: 
Image

Project assets: 
Image

zateutsch:
Fixes #51 

Just made it so sample asset files are copied over on project export.

@azchohfi should probably look at this one since I was messing with the generator stuff.

### DIFF
--- a/AIDevGallery/ProjectGenerator/Generator.cs
+++ b/AIDevGallery/ProjectGenerator/Generator.cs
@@ -267,6 +267,17 @@ internal partial class Generator
             }
         }
 
+        // Add Asset Files
+        foreach(string assetFilename in sample.AssetFilenames)
+        {
+            string fullAssetPath = Path.Join(Windows.ApplicationModel.Package.Current.InstalledLocation.Path, "Assets", assetFilename);
+            string fullOutputAssetPath = Path.Combine(outputPath, "Assets", assetFilename);
+            if (Path.GetExtension(fullAssetPath) is ".jpg" or ".png")
+            {
+                File.Copy(fullAssetPath, fullOutputAssetPath);
+            }
+        }
+
         var csproj = Path.Join(outputPath, $"{safeProjectName}.csproj");
 
         // Add NuGet references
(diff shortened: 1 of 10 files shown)


AIDevGallery/Samples/Open Source Models/Image Models/Faster RCNN/ObjectDetection.xaml.cs
```diff
@@ -30,6 +30,9 @@ namespace AIDevGallery.Samples.OpenSourceModels.ObjectDetection.FasterRCNN;
         "Microsoft.ML.OnnxRuntime.DirectML",
         "Microsoft.ML.OnnxRuntime.Extensions"
     ],
+    AssetFilenames = [
+        "team.jpg"
```
zateutsch:
```suggestion
        "pose_default.png"
```

AIDevGallery/ProjectGenerator/Generator.cs
```diff
@@ -267,6 +267,17 @@ private async Task<string> GenerateAsyncInternal(Sample sample, Dictionary<Model
             }
         }
 
+        // Add Asset Files
+        foreach(string assetFilename in sample.AssetFilenames)
+        {
+            string fullAssetPath = Path.Join(Windows.ApplicationModel.Package.Current.InstalledLocation.Path, "Assets", assetFilename);
```
azchohfi:
```suggestion
            string fullAssetPath = Path.Join(Package.Current.InstalledLocation.Path, "Assets", assetFilename);
```

azchohfi (approved):
I love this!
My only suggestion is to not use strings, but rather do the equivalent to what we are already doing for the SharedCode enum. So you could have a generator that generates an enum for each/all Assets.
Still, I'm ok merging this as is.

PR #96, opened 2025-01-19
#83: Added reset to default button to custom params sample

### LINKED ISSUE #83: [FEATURE] A "Reset to default" button would be nice in the custom parameter sample
**Is your feature request related to a problem? Please describe.**
I was messing around with the parameters in the Custom Parameter sample and ended up getting gibberish after messing with the sliders. I was curious how to get back to output that makes sense but didn't remember the default values. 

**Describe the solution you'd like**
A reset to default values button.

**Describe alternatives you've considered**
Reloading the sample completely but this isn't good UX. Leaving the sample and then coming back.

**Additional context**

zateutsch:
fixes #83 

Reset to default button added to custom parameters sample.


AIDevGallery/Samples/SharedCode/GenAIModel.cs
```diff
@@ -19,12 +19,12 @@ internal class GenAIModel : IChatClient, IDisposable
     private const string TEMPLATE_PLACEHOLDER = "{{CONTENT}}";
 
     // Search Options
-    private const int DefaultTopK = 50;
-    private const float DefaultTopP = 0.9f;
-    private const float DefaultTemperature = 1;
-    private const int DefaultMinLength = 0;
+    public const int DefaultTopK = 50;
```
azchohfi: We should move away from having these as public, not making them all public. We should not rely on GenAIModel class for these values. Ideally, these are provided to the sample, or recovered in a generic way, since different implementations of IChatClient will have different defaults, and we want the sample to work with all of them.

azchohfi (changes requested):

nmetulev (approved):

PR #98, opened 2025-01-20
#97: Adjusting font size for each prediction box

### LINKED ISSUE #97: [ENHANCEMENT] Prevent labels from exceeding prediction box on Detect Objects
Currently predictions look like this and go out of bounds:

Image

Think they should be bounded inside the box except for when the font size would be too small.

zateutsch:
fixes #97 

Labels were going outside of boxes for predictions. Now: calculating font size and clamping it so can't be too big or too small, but will usually stay within prediction bounds

Also changed the default image because it was hard to see the labels on the default.


AIDevGallery/Samples/SharedCode/BitmapFunctions.cs
```diff
@@ -220,8 +216,11 @@ public static BitmapImage RenderPredictions(Bitmap image, List<Prediction> predi
             g.DrawLine(pen, p.Box.Xmax, p.Box.Ymax, p.Box.Xmin, p.Box.Ymax);
             g.DrawLine(pen, p.Box.Xmin, p.Box.Ymax, p.Box.Xmin, p.Box.Ymin);
 
-            // Draw the label and confidence
+            // Calculate adjusted font and  draw the label
+            float maxTextWidth = p.Box.Xmax - p.Box.Xmin;
             string labelText = $"{p.Label}, {p.Confidence:0.00}";
+            float adjustedFontSize = Math.Clamp(maxTextWidth / ((float)labelText.Length), 8, 16);
+            using Font font = new("Arial", adjustedFontSize);
```
azchohfi: I would move definition outside of the loop, so we create only one reference, if possible.
I see the logic using the value from the specific prediction, but its already clamped between 8 and 16. Couldn't we pick a value, or maybe the average between all predictions? Something like that. For example, select the team.jpg file we have on the gallery Assets. It renders each prediction with a slightly different size, which feels weird.

nmetulev (approved):

PR #100, opened 2025-01-20
Limits minimum temperature in Custom parameters to 0.01

### LINKED ISSUE #99: [BUG] Custom parameters generates garbage when Temperature is set to 0
**Describe the bug**
When you set the temperature in Custom Parameters to 0, it will generate garbage

**To Reproduce**
1. Go to Custom Parameters sample
2. Move the Temperature slider to 0
3. Type a prompt
4. Click on Generate
5. See error

**Expected behavior**
There should be a response to the prompt

**Screenshots**
Image

**Additional context**
This is due to the source code in the onnx runtime that scales the score by the temperature:
https://github.com/microsoft/onnxruntime-genai/blob/a6bec83bf02667f6a9c6aa6340842f46e001578c/src/search.cpp#L147 

A temperature of 0 will generate a division by 0

brunosonnino:
Fixes #99

### DIFF
--- a/AIDevGallery/Samples/Open Source Models/Language Models/CustomSystemPrompt.xaml
+++ b/AIDevGallery/Samples/Open Source Models/Language Models/CustomSystemPrompt.xaml
@@ -108,7 +108,7 @@
                         FontSize="12"
                         Header="Temperature"
                         Maximum="5"
-                        Minimum="0"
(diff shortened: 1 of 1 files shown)


zateutsch (commented):
lgtm

PR #117, opened 2025-01-23
Upgrading to Windows App SDK 1.7-exp2

niels9001:
- Upgrading to WASDK1.7-exp2
- Fixing a but where the NavView buttons were left-aligned due to this upgrade.
- Swapping out our custom titlebar with the new WinUI titlebar


AIDevGallery/Styles/NavigationView.xaml
```diff
@@ -345,7 +345,7 @@
                             <SplitView.Pane>
                                 <Grid
                                     x:Name="PaneContentGrid"
-                                    HorizontalAlignment="Left"
+                                    HorizontalAlignment="Stretch"
```
azchohfi, line 348: Are we still using all the resources in this resource dictionary?

nmetulev (approved):

AIDevGallery/Styles/NavigationView.xaml
nmetulev: Yes

PR #120, opened 2025-01-23
Minor tweaks to search

nmetulev:
* Fixed "no results" message to show up in cases where there are results from the api but none that work for the app
* Added button to each results to open the model card in the browser
* Added missing license


niels9001 (approved):
LGTM! Removed the Guides page and well and tweaked the icon a bit

PR #123, opened 2025-01-24
#106: Save Custom Parameters State with AppData

### LINKED ISSUE #106: [SAMPLE] Custom Parameters samples should remember state
Is this a sample for an existing model or API, or for a new model? (If a new model, provide a link to the model in ONNX format)
Yes, for the Custom Parameters sample for GenAI Models/APIs

Describe the scenarios(s)
Whenever I switch between models, i have to reenter my system prompt and user prompt which makes it harder to test. We should keep the state of sample (including parameters) between reloads.

zateutsch:
fixes #106 

Custom parameters will preserve last state with AppData class.


niels9001:
@zateutsch FYI, I misued your PR for the single UI column:

image

nmetulev (approved):

PR #127, opened 2025-01-24
Simplified exported projects.

### LINKED ISSUE #32: [FEATURE] Simplify exported projects
**Is your feature request related to a problem? Please describe.**
The exported projects for each sample contain too much of the gallery shell and should be simplified even further to optimize for simplicity and getting started. The current papercuts of the exported projects are:

- The page to the sample is in a different file than the MainWindow where I would expect it. The file is also named differently for each sample, making it difficult to find exactly which one it is
- The actual configuration of the sample is all contained in the SampleNavigationParameters.cs and that is not intuitive
- Namespaces for projects are different making it harder to mix samples together

**Describe the solution you'd like**
We should change how the samples are exported by:

- [x] We should remove the SampleNavigationParameters and inject the values in the MainWIndow class as properties, or directly in the OnNavigatedTo method
- [x] We should replace SampleNavigationParameter.RequestWaitForCompletion and SampleNavigationParameter.NotifyCompletion with code to show the injected loading view. We can pattern match the line if (e.Parameter is SampleNavigationParameters sampleParam) to identify the name of the parameter and replace usage accordingly.
- [x] Helper classes should be in a separate folder to make it easy to identify the entry point to the sample (MainWindow)
- [x] Make the root namespace the same for all projects. AIDevGallery.Sample and AIDevGallery.Sample.Utils
- [x] Create sln file
- [ ] Rename the sample page to sample.xaml and sample.xaml.cs to be consistent across samples and easier to find.

azchohfi:
Fixes #32.

### DIFF
--- a/AIDevGallery.UnitTests/ProjectGeneratorUnitTests.cs
+++ b/AIDevGallery.UnitTests/ProjectGeneratorUnitTests.cs
@@ -167,7 +167,7 @@ public class ProjectGenerator
         {
             FileName = @"C:\Program Files\dotnet\dotnet",
             WorkingDirectory = projectPath,
-            Arguments = $"build -r win-{arch} -f {Generator.DotNetVersion}-windows10.0.22621.0 /p:Configuration=Release /p:Platform={arch} /flp:logfile={logFileName}",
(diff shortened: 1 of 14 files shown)


nmetulev:
Works great. The only suggestion - we should rename the sample page to sample.xaml and sample.xaml.cs to be consistent across samples and easier to find.

azchohfi:
> Works great. The only suggestion - we should rename the sample page to sample.xaml and sample.xaml.cs to be consistent across samples and easier to find.

Done

nmetulev (approved):

PR #145, opened 2025-01-31
Scenarios overview page

niels9001:
image


nmetulev (commented):
I like it.

Should we have the npu/gpu filter in the page also?

nmetulev:
Also, the back button in general is messed up with this flow.

nmetulev (changes requested):
Amazing!

Before merging:
1. Remove Action Button - unless there is something specific we need there
1. Should we now always navigate to overview page instead of last sample?
2. Handle narrow view 
image
3. Should there be a filter for picking CPU/GPU/NPU/API?

brunosonnino (approved):

PR #154, opened 2025-02-05
#118: Parse out markdown syntax before and after generated code

### LINKED ISSUE #118: [BUG] Generate code output includes markdown syntax
This is minor, but I want to parse this (```csharp) out in real time if possible:

Image

I have had no luck with prompting to get it to stop and I want it gone now.

zateutsch:
fixes #118 

Parse out markdown syntax before and after code live during generation, this should also work fine if there is no markdown, so should cover all our bases.


nmetulev (commented):
I'm seeing markdown still in the response. This is with Phi 3.5 
image

nmetulev (approved):

PR #155, opened 2025-02-06
#113: Semantic Kernel

### LINKED ISSUE #113: [SAMPLE] Create a Semantic Kernel Chat with Onnx
### Is this a sample for an existing model or API, or for a new model? (If a new model, provide a link to the model in ONNX format)
Yes, it's a sample for a chat application with Semantic Kernel

### Describe the scenarios(s)
It's a chat application using the Semantic Kernel API to handle history and streaming response

zateutsch:
fixes #113 

Needed to add a dependency and update another for this, so would appreciate a double check that nothing went awry.

Also, this sample takes *forever* to load. Not sure if there is any way around it.


zateutsch:
@nmetulev this sample has some problems, I'm investigating

azchohfi:
@zateutsch I've fixed the sample with a more generic solution (using IChatClient's overload). This will also work well will PhiSilica, so its only a plus :) The package you were referencing have its own implementation of the equivalent of IChatClient for ORT, so we should not use it (it was fixed to the CPU version).

azchohfi (approved):

zateutsch:
I don't think we should merge this until I've double checked that that memory leak is gone

AIDevGallery/Samples/Open Source Models/Language Models/SemanticKernelChat.xaml.cs
```diff
@@ -0,0 +1,267 @@
...
+    NugetPackageReferences = [
+        "CommunityToolkit.Mvvm",
+        "Microsoft.ML.OnnxRuntimeGenAI.DirectML",
+        "Microsoft.SemanticKernel.Core"
+    ],
+    SharedCode = [
+        SharedCodeEnum.GenAIModel,
+        SharedCodeEnum.ChatTemplateSelector,
+        SharedCodeEnum.Message,
+    ])]
```
nmetulev: Updating to match recent changes to how language samples are added

```suggestion
    NugetPackageReferences = [
        "CommunityToolkit.Mvvm",
        "Microsoft.SemanticKernel.Core"
    ],
    SharedCode = [
        SharedCodeEnum.ChatTemplateSelector,
        SharedCodeEnum.Message,
    ])]
```

nmetulev (approved):

Issue #166, opened 2025-02-07
[BUG] Remove background sample says you can select 32 points but crashes if you do


Jaylyn-Barbee:
**Describe the bug**
If you select 32 points in the remove background sample it crashes.

**To Reproduce**

1. Open the gallery and nav to the remove background sample
2. Load image
3. Click 32 points
4. Observe crash

**Expected behavior**
Either 32 points should work or we should change to 31

**Screenshots**
![Image](https://github.com/user-attachments/assets/21e8ce1c-6671-4696-9e7a-d5cac1885393)

**Please complete the following information:**
Branch: alzollin/wcrapis

**Additional context**

PR #177, opened 2025-02-10
Lowered the max points for WCR Remove background

Jaylyn-Barbee:
Fixes: #166

Lowered the max points from 32 to 31 because 32 causes a crash.


zateutsch (approved):
lgtm

PR #178, opened 2025-02-10
Added sample interaction telemetry event.

azchohfi:
TODO: add to other samples after initial review.

### DIFF
--- a/AIDevGallery/Samples/BaseSamplePage.cs
+++ b/AIDevGallery/Samples/BaseSamplePage.cs
@@ -38,7 +38,7 @@ internal partial class BaseSamplePage : Page
         return Task.CompletedTask;
     }
 
-    internal void SampleInteractedEvent(string? customInfo = null)
+    internal void SendSampleInteractedEvent(string? customInfo = null)
     {
         SampleParams?.SendSampleInteractionEvent(customInfo);
     }
--- a/AIDevGallery/Samples/BaseSamplePage.cs
+++ b/AIDevGallery/Samples/BaseSamplePage.cs
@@ -10,16 +10,20 @@ namespace AIDevGallery.Samples;
 
 internal partial class BaseSamplePage : Page
 {
+    private BaseSampleNavigationParameters? SampleParams { get; set; }
+
     protected override async void OnNavigatedTo(NavigationEventArgs e)
     {
         base.OnNavigatedTo(e);
 
         if (e.Parameter is SampleNavigationParameters sampleParams)
         {
+            SampleParams = sampleParams;
             await LoadModelAsync(sampleParams);
         }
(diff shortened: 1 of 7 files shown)


AIDevGallery/Samples/BaseSamplePage.cs
```diff
@@ -33,4 +37,9 @@ protected virtual Task LoadModelAsync(MultiModelSampleNavigationParameters sampl
     {
         return Task.CompletedTask;
     }
+
+    internal void SampleInteractedEvent(string? customInfo = null)
```
nmetulev:
```suggestion
    internal void SendSampleInteractedEvent(string? customInfo = null)
```

AIDevGallery/Samples/Open Source Models/Language Models/Chat.xaml.cs
```diff
@@ -122,6 +122,7 @@ private void AddMessage(string text)
         Messages.Add(new Message(text.Trim(), DateTime.Now, ChatRole.User));
         var contentStartedBeingGenerated = false; // <exclude-line>
         NarratorHelper.Announce(InputBox, "Generating response, please wait.", "ChatWaitAnnouncementActivityId"); // <exclude-line>>
+        SampleInteractedEvent(); // <exclude-line>
```
nmetulev:
```suggestion
        SampleInteractedEvent("AddMessage"); // <exclude-line>
```

AIDevGallery/Telemetry/Events/SampleInteractionEvent.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace AIDevGallery.Telemetry.Events;
+
+[EventData]
+internal class SampleInteractionEvent : EventBase
+{
+    public override PartA_PrivTags PartA_PrivTags => PrivTags.ProductAndServiceUsage;
+
+    public string SampleId { get; }
+    public string Model1Id { get; }
```
nmetulev, line 18: why not make a models array instead of hardcoding model1, model2, etc?

nmetulev: Also, since these are not just models - we should think about how we make this more readable and move away from just calling everything model

nmetulev (changes requested):

AIDevGallery/Telemetry/Events/SampleInteractionEvent.cs
azchohfi: These are reflected by the telemetry [EventData], so we should be very clear with what we call them.
We do not have plans on having samples with more than two models/endpoints (only the rag and multipose samples uses this). Adding more than two would make the UI very complex too.

PR #194, opened 2025-02-12
Updates to how WCR APIs show up in the app

### LINKED ISSUE #165: [BUG] View model card button shown when WCR API is disabled
Image

There shouldn't be any licenses and modelcards in this view

### LINKED ISSUE #168: [BUG] [WCR] APIs should always be on top of model picker if available
APIs should always be on top of model picker if available

### LINKED ISSUE #170: [BUG] [WCR] APIs should show as unavailable on non supported machines
WCR APIs should show as unavailable on non supported machines. Right now they are available and the samples can be loaded but don't do anything

### LINKED ISSUE #173: [BUG] [WCR] API Compatibility warning should be more specific
**Describe the bug**
Right now, the APIs are checking for ARM device and that is not enough. 

Image

**Expected behavior**
We should check if the machine is a Copilot PC+ and identify when it's not to let the user know then they need one
We should also check if the machine is on a Windows build where the APIs are supported and let the user when they are not

### LINKED ISSUE #174: [BUG] [WCR] APIs should not request the model to be download automatically and give the user a choice to request the download
**Describe the bug**
Right now, when opening a sample with Phi Silica, the model is requested to be downloaded automatically. Some users might not want to do that, especially if we make phi silica the first option on the list.

nmetulev:
Fixes #168 
Fixes #170
Fixes #173 
Fixes #174 
Fixes #165


AIDevGallery.SourceGenerator/SamplesSourceGenerator.cs
```diff
@@ -66,7 +66,15 @@ private void ExecuteSharedCodeEnumGeneration(SourceProductionContext context, Im
...
+            var filePathWithoutExtension = filePath.Substring(0, filePath.Length - extension.Length);
+            if (fileName.EndsWith(".xaml", StringComparison.InvariantCultureIgnoreCase) && File.Exists(filePathWithoutExtension))
+            {
+                fileName = Path.GetFileNameWithoutExtension(filePathWithoutExtension);
+                sourceBuilder.AppendLine($"        {fileName}Cs,");
+                sourceBuilder.AppendLine($"        {fileName}Xaml,");
+            }
+            else if (File.Exists(Path.ChangeExtension(filePath, ".xaml")))
             {
                 sourceBuilder.AppendLine($"        {fileName}Cs,");
```
azchohfi, line 79: Is this ever being called now? I mean for xaml files and not xaml.cs files? If its not, we could remove this.

AIDevGallery/Controls/ModelSelectionControl.xaml.cs
```diff
@@ -265,9 +265,24 @@ private void PopulateModelDetailsLists()
...
     }
 
+    private void SortModels<T>(ObservableCollection<T> models)
+        where T : IModelView
+    {
+        var sortedModels = models.OrderByDescending(m => m.ModelDetails.HardwareAccelerators.Contains(HardwareAccelerator.WCRAPI)).ToList();
+        models.Clear();
+        foreach (var model in sortedModels)
+        {
+            models.Add(model);
```
azchohfi: Doing a clear and a bunch of adds will trigger multiple UI updates (INCC), which literally just happened on the method that calls this method.
It would be better to use regular lists on the PopulateModelDetailsLists method, and add only once to the OC.

AIDevGallery.SourceGenerator/Models/HardwareAccelerator.cs
```diff
@@ -10,5 +10,6 @@ internal enum HardwareAccelerator
 {
     CPU,
     DML,
-    QNN
+    QNN,
+    WCRAPI
```
azchohfi, line 14: Is this logic specific for WCRAPIs or for all future APIs (Azure, OAI, etc)?

AIDevGallery/Models/ModelCompatibility.cs
```diff
@@ -20,9 +20,24 @@ private ModelCompatibility()
...
     {
         string description = string.Empty;
-
         ModelCompatibilityState compatibility;
-        if (modelDetails.HardwareAccelerators.Contains(HardwareAccelerator.CPU) ||
+
+        // check if WCR API
+        if (modelDetails.Url.StartsWith("file://", StringComparison.InvariantCultureIgnoreCase))
+        {
+            var apiKey = modelDetails.Url.Substring(7);
```
azchohfi: The if should use the underlying API information:
```cs
        if (Samples.ModelTypeHelpers.ApiDefinitionDetails.Any(md => md.Value.Id == modelDetails.Id))
```
And you can retrieve the key without using Enum.Parse/TryParse:
```cs
        var apiKey = Samples.ModelTypeHelpers.ApiDefinitionDetails.FirstOrDefault(md => md.Value.Id == modelDetails.Id).Key;
```
This should be more reliable, and work for all APIs.

AIDevGallery/Pages/APISelectionPage.xaml.cs
```diff
@@ -25,7 +26,13 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
         {
             if (e.Parameter is ModelType type)
             {
-                SetSelectedAPIInMenu(type);
+                SetSelectedApiInMenu(type);
+            }
+            else if (e.Parameter is ModelDetails details &&
+                    details.Url.StartsWith("file://", StringComparison.InvariantCultureIgnoreCase) &&
+                    Enum.TryParse(details.Url.Substring(7), out ModelType apiType))
```
azchohfi: Same here.

AIDevGallery/ProjectGenerator/Generator.cs
```diff
@@ -674,10 +674,16 @@ private string CleanXamlSource(string xamlCode, string newNamespace, out string
         if (match.Success)
         {
             var oldClassFullName = match.Groups[1].Value;
-            _ = oldClassFullName[..oldClassFullName.LastIndexOf('.')];
             className = oldClassFullName[(oldClassFullName.LastIndexOf('.') + 1)..];
 
-            xamlCode = xamlCode.Replace(match.Value, @$"x:Class=""{newNamespace}.Sample""");
+            if (oldClassFullName.Contains(".SharedCode."))
+            {
+                xamlCode = xamlCode.Replace(match.Value, @$"x:Class=""{newNamespace}.{className}""");
```
azchohfi, line 681: What does this do precisely? Is it because of the rename/move?

azchohfi (changes requested):

AIDevGallery.SourceGenerator/SamplesSourceGenerator.cs
nmetulev: Yes it is.

AIDevGallery.SourceGenerator/Models/HardwareAccelerator.cs
nmetulev: just for wcrapis

AIDevGallery/Controls/ModelSelectionControl.xaml.cs
nmetulev: Removed sorting and just inserting apis on top now

AIDevGallery/ProjectGenerator/Generator.cs
nmetulev: done

AIDevGallery/Models/ModelCompatibility.cs
nmetulev: done

AIDevGallery/Pages/APISelectionPage.xaml.cs
nmetulev: done

AIDevGallery/ProjectGenerator/Generator.cs
nmetulev: There is a bug when a usercontrol is exported, the class is renamed to .Sample instead of the usercontrol class name. This fixed that

AIDevGallery/Models/ModelCompatibility.cs
```diff
@@ -20,9 +21,23 @@ private ModelCompatibility()
     public static ModelCompatibility GetModelCompatibility(ModelDetails modelDetails)
     {
         string description = string.Empty;
-
         ModelCompatibilityState compatibility;
-        if (modelDetails.HardwareAccelerators.Contains(HardwareAccelerator.CPU) ||
+
+        // check if WCR API
+        if (modelDetails.Url.StartsWith("file://", StringComparison.InvariantCultureIgnoreCase))
```
azchohfi: This check is doing the exact same thing as the next check (the .Any). Checking for Url to start with file:// is flaky, and we might change this. Check if it is an API by searching for the ID, like your next if statement only.

AIDevGallery/Pages/APISelectionPage.xaml.cs
```diff
@@ -25,7 +27,13 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
         {
             if (e.Parameter is ModelType type)
             {
-                SetSelectedAPIInMenu(type);
+                SetSelectedApiInMenu(type);
+            }
+            else if (e.Parameter is ModelDetails details &&
+                    details.Url.StartsWith("file://", StringComparison.InvariantCultureIgnoreCase) &&
```
azchohfi: You don't need both checks here neither.

nmetulev: Done

AIDevGallery/Models/ModelCompatibility.cs
nmetulev: Done

azchohfi (approved):

PR #196, opened 2025-02-12
Added interaction events on all samples.

azchohfi:
Follow up to #178

### DIFF
--- a/AIDevGallery/Samples/Open Source Models/Embeddings/SemanticSearch.xaml.cs
+++ b/AIDevGallery/Samples/Open Source Models/Embeddings/SemanticSearch.xaml.cs
@@ -129,6 +129,8 @@ internal sealed partial class SemanticSearch : BaseSamplePage
             return;
         }
 
+        SendSampleInteractedEvent("Search"); // <exclude-line>
+
         Task.Run(
             async () =>
             {
--- a/AIDevGallery/Samples/SharedCode/SemanticComboBox.cs
+++ b/AIDevGallery/Samples/SharedCode/SemanticComboBox.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using AIDevGallery.Telemetry.Events;
 using Microsoft.Extensions.AI;
 using Microsoft.Extensions.VectorData;
 using Microsoft.SemanticKernel.Connectors.InMemory;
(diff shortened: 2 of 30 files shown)


AIDevGallery/Samples/Open Source Models/Embeddings/SemanticSearch.xaml.cs
```diff
@@ -129,6 +129,8 @@ public void Search(string sourceText, string searchText)
             return;
         }
 
+        SendSampleInteractedEvent("Search"); // <exclude-line>
```
nmetulev, line 132: We should auto exclude lines that call SendSampleInteractedEvent to avoid having to add exclude-line

AIDevGallery/Samples/SharedCode/SemanticComboBox.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using AIDevGallery.Telemetry.Events;
```
nmetulev, line 4: should we also exclude-line these usings?

azchohfi: I don't think we have to, since the exported projects are building fine (tests are passing). I believe these are already being removed from the SharedCode source generator.

AIDevGallery/Samples/Open Source Models/Embeddings/SemanticSearch.xaml.cs
azchohfi: I didn't want to add yet another magical exclude pattern.
Most of these are in blocks that are already being removed, like `NarratorHelper.Announce(...) // <exclude-line>`

nmetulev: these should never make it to the exported samples - lets just make sure they don't end up in sample code

AIDevGallery/Samples/SharedCode/SemanticComboBox.cs
nmetulev: they show up in sample code 
image

azchohfi: Oh, interesting. This doesn't show up in the exported project itself. This is probably being filtered at the wrong moment. It needs to happen on the SourceGenerator, so during compilation, not project export. I'll fix this.

azchohfi: Done.

AIDevGallery/Samples/Open Source Models/Embeddings/SemanticSearch.xaml.cs
azchohfi: Done.

nmetulev (approved):

PR #202, opened 2025-02-13
#184: Fix relative WCR readme links

### LINKED ISSUE #184: [BUG] [WCR] Clicking links in API description throws an exception because the links are relative
The links in the API description page should work and not throw an exception. If the link is relative, it should still work.

zateutsch:
fixes #184 

I believe this captures all types of links in WCR docs

### DIFF
--- a/AIDevGallery/Pages/ModelPage.xaml.cs
+++ b/AIDevGallery/Pages/ModelPage.xaml.cs
@@ -12,6 +12,7 @@ using Microsoft.UI.Xaml.Navigation;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


AIDevGallery/Pages/ModelPage.xaml.cs
```diff
@@ -219,4 +229,27 @@ private void SampleList_ItemInvoked(ItemsView sender, ItemsViewItemInvokedEventA
...
+        if(link.StartsWith('/'))
+        {
+            fixedLink = Path.Join(DocsBaseUrl, link);
+        }
+        else
+        {
+            fixedLink = Path.Join(DocsBaseUrl, WcrDocsRelativePath, link.Replace(".md", string.Empty));
+        }
+
+        System.Diagnostics.Debug.WriteLine(fixedLink);
```
zateutsch:
```suggestion
```

AIDevGallery/Pages/ModelPage.xaml.cs
```diff
@@ -21,6 +22,8 @@ namespace AIDevGallery.Pages;
 
 internal sealed partial class ModelPage : Page
 {
+    private const string DocsBaseUrl = "https://learn.microsoft.com/en-us/";
```
nmetulev:
```suggestion
    private const string DocsBaseUrl = "https://learn.microsoft.com/";
```

Should work without the localized link so it can redirect to current user local

nmetulev (approved):

Issue #189, opened 2025-02-11
[a11y BUG] Narrator focus is moving to decorative image and announcing as Image in the Generate Text Page.


gregwoo-microsoft:
Repro steps:
Launch AI Dev Gallery App
Navigate to Scenarios button in the left pane and invoke it
Navigate to Text Dropdown and invoke it
Navigate to Generate text and invoke it
Now Navigate to Decorative image, which is available next to Pho 3 Mini DirectML label and observe the behavior

We had fixed this issue in the past, but it still repros here:
Samples->Image->Generate Code.
Samples->Images->Detect Human Pose
Samples->Images->Detect Multiple Human Pose
Enhance Images
Segment Street

![Image](https://github.com/user-attachments/assets/0763205e-a7b5-411a-be9a-18a13c193938)

PR #219, opened 2025-02-19
Icon has an a11y name, fixes a11y bug

gregwoo-microsoft:
#189 

image

### DIFF
--- a/AIDevGallery/Controls/ModelSelectionControl.xaml
+++ b/AIDevGallery/Controls/ModelSelectionControl.xaml
@@ -101,7 +101,7 @@
                                         </Grid>
                                     </toolkit:SettingsCard.Header>
                                     <toolkit:SettingsCard.HeaderIcon>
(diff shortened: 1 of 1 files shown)


niels9001:
@gregwoo-microsoft Have we heard back from the a11y team what the expected behavior should be? 

This is a decorative image, so as far as I'm aware it should not have a name and should not be labeled as `Content` or `Control`

From the WinUI 3 Gallery:

image

gregwoo-microsoft:
My understanding is that their concern came from the fact that the image was still accessible via scan mode and called as "image". My change fixes that by providing a more descriptive name. But I agree with you that setting it to "raw" should be a preferred solution. I've reached out to them to confirm. I'll let you know once I hear back from them.

niels9001:
> My understanding is that their concern came from the fact that the image was still accessible via scan mode and called as "image". My change fixes that by providing a more descriptive name. But I agree with you that setting it to "raw" should be a preferred solution. I've reached out to them to confirm. I'll let you know once I hear back from them.

Yeah.. maybe we set it to Raw AND set the AutomationProperties.Name? Wonder if that works..

PR #222, opened 2025-02-20
Changes in the prompt system for Phi-Silica

### LINKED ISSUE #179: [BUG] [WCR] Phi Silica output contains role/stop tokens
It doesn't seem to always do this, but sometimes:

Image

Not sure what the cause is, probably just a bug with Silica, but worth checking if it is something on our end with prompts?

brunosonnino:
Fixes #179


AIDevGallery/Samples/SharedCode/WcrApis/PhiSilicaClient.cs
```diff
@@ -191,51 +183,19 @@ private string GetPrompt(IEnumerable<ChatMessage> history)
...
 
-                if (string.IsNullOrWhiteSpace(_promptTemplate.System))
-                {
-                    systemMsgWithoutSystemTemplate = message.Text ?? string.Empty;
-                }
-                else
-                {
-                    prompt += _promptTemplate.System.Replace(TEMPLATE_PLACEHOLDER, message.Text);
-                }
+                _languageModelContext = _languageModel?.CreateContext(message.Text, new ContentFilterOptions());
```
azchohfi: This assignment is overwriting the previous one. I believe it would be better to create this only once. I believe you could not do the first one if the first message 's role is system.

AIDevGallery/Samples/SharedCode/WcrApis/PhiSilicaClient.cs
```diff
@@ -328,6 +289,8 @@ public async IAsyncEnumerable<string> GenerateStreamResponseAsync(string prompt,
 
             var response = await progress;
 
+            _languageModelContext = null;
```
azchohfi: Why do you need to set it to null at every new inference?

AIDevGallery/Samples/SharedCode/WcrApis/PhiSilicaClient.cs
```diff
@@ -295,14 +255,15 @@ public async IAsyncEnumerable<string> GenerateStreamResponseAsync(string prompt,
         if (!_languageModel.IsPromptLargerThanContext(prompt))
         {
             IAsyncOperationWithProgress<LanguageModelResponse, string>? progress;
+            LanguageModelContext context = _languageModelContext ?? _languageModel.CreateContext();
```
azchohfi: Actually, since this is lazily created, just remove the first creation.

AIDevGallery/Samples/SharedCode/WcrApis/PhiSilicaClient.cs
brunosonnino: I refactored the creation, now it's done only once

AIDevGallery/Samples/SharedCode/WcrApis/PhiSilicaClient.cs
brunosonnino: This was an excess of care - removed

AIDevGallery/Samples/SharedCode/WcrApis/PhiSilicaClient.cs
brunosonnino: I removed the creation here - now, it's always created in GetPrompt

azchohfi (approved):

PR #224, opened 2025-02-20
[WCR] Describe image xaml cleanup

zateutsch:


nmetulev:
Shold we actually have the description and loader go on top of the image? This way, the image can go full height.

PR #236, opened 2025-02-24
Sample Container - Fix for concurrency bug

nmetulev:
Fixes an issue when a sample is still loading and the user navigates to a different sample in the same sample container, the previous sample could break current sample loading.


AIDevGallery/Controls/SampleContainer.xaml.cs
```diff
@@ -239,8 +242,11 @@ public async Task LoadSampleAsync(Sample? sample, List<ModelDetails>? models)
 
         await _sampleLoadedCompletionSource.Task;
```
azchohfi, line 243:
```suggestion
        await _sampleLoadedCompletionSource.Task.WaitAsync(token);
```

AIDevGallery/Controls/SampleContainer.xaml.cs
```diff
@@ -107,6 +107,10 @@ public SampleContainer()
 
     public async Task LoadSampleAsync(Sample? sample, List<ModelDetails>? models)
     {
+        CancelCTS();
+        _sampleLoadingCts = new CancellationTokenSource();
+        var currentLoadingCts = _sampleLoadingCts;
+
```
azchohfi, line 113: Keep the CancelCTS where it was. That was the right point to do this.

AIDevGallery/Controls/SampleContainer.xaml.cs
```diff
@@ -178,7 +180,9 @@ public async Task LoadSampleAsync(Sample? sample, List<ModelDetails>? models)
                     SampleFrame.Content = null;
 
                     VisualStateManager.GoToState(this, "WcrModelNeedsDownload", true);
-                    if (!await modelDownloader.SetDownloadOperation(ModelType.PhiSilica, sample.Id, LanguageModel.MakeAvailableAsync))
+                    if (!await modelDownloader.SetDownloadOperation(ModelType.PhiSilica, sample.Id, LanguageModel.MakeAvailableAsync) ||
```
azchohfi, line 183: Bring the lines:
```cs
_sampleLoadingCts = new CancellationTokenSource();
var token = _sampleLoadingCts.Token;
```
to before the `// if PhiSilica, only show model loader and reload sample once loaded`

Then change this line to:
```suggestion
                    if (!await modelDownloader.SetDownloadOperation(ModelType.PhiSilica, sample.Id, LanguageModel.MakeAvailableAsync).WaitAsync(token) ||
```

AIDevGallery/Controls/SampleContainer.xaml.cs
```diff
@@ -196,7 +200,6 @@ public async Task LoadSampleAsync(Sample? sample, List<ModelDetails>? models)
         VisualStateManager.GoToState(this, "SampleLoading", true);
         SampleFrame.Content = null;
 
-        _sampleLoadingCts = new CancellationTokenSource();
         _sampleLoadedCompletionSource = new TaskCompletionSource();
         BaseSampleNavigationParameters sampleNavigationParameters;
```
azchohfi, line 205: Delete `var token = _sampleLoadingCts.Token;` from here, as you moved it up.

AIDevGallery/Controls/SampleContainer.xaml.cs
azchohfi: You also need to wrap this call in a try catch, and handle the OperationCanceledException.

azchohfi (changes requested):

AIDevGallery/Controls/SampleContainer.xaml.cs
```diff
@@ -239,8 +242,11 @@ public async Task LoadSampleAsync(Sample? sample, List<ModelDetails>? models)
 
         await _sampleLoadedCompletionSource.Task;
 
-        _sampleLoadedCompletionSource = null;
-        _sampleLoadingCts = null;
+        if (currentLoadingCts != _sampleLoadingCts || currentLoadingCts.IsCancellationRequested)
+        {
+            // we are no longer in the same sample
+            return;
+        }
```
azchohfi, line 249: Delete this:
```suggestion
```

AIDevGallery/Controls/SampleContainer.xaml.cs
```diff
@@ -407,11 +413,15 @@ private async void WcrModelDownloader_DownloadClicked(object sender, EventArgs e
...
+            var model = _modelsCache;
+            var currentCts = _sampleLoadingCts;
+
             var op = LanguageModel.MakeAvailableAsync();
-            if (await modelDownloader.SetDownloadOperation(op))
+            if (await modelDownloader.SetDownloadOperation(op) && currentCts == _sampleLoadingCts)
             {
                 // reload sample
-                _ = this.LoadSampleAsync(_sampleCache, _modelsCache);
+                _ = this.LoadSampleAsync(sample, model);
```
azchohfi, line 424: Revert this.

azchohfi (changes requested):

AIDevGallery/Controls/SampleContainer.xaml.cs
azchohfi: For example:
```cs
        try
        {
            await _sampleLoadedCompletionSource.Task.WaitAsync(token);
        }
        catch (OperationCanceledException)
        {
            return;
        }
        finally
        {
            _sampleLoadedCompletionSource = null;
            _sampleLoadingCts = null;
        }
```

PR #243, opened 2025-02-25
Add stop button to Image Description.

### LINKED ISSUE #239: [BUG] WCR image description should have a stop button
We should have a stop button in the Image Description WCR sample

Image

brunosonnino:
__Note__: Image Description cannot be canceled before the first word received 
Fixes #239

### DIFF
--- a/AIDevGallery/Samples/WCRAPIs/ImageDescription.xaml
+++ b/AIDevGallery/Samples/WCRAPIs/ImageDescription.xaml
@@ -56,6 +56,7 @@
                 <Grid.ColumnDefinitions>
                     <ColumnDefinition Width="Auto" />
                     <ColumnDefinition Width="*" />
+                    <ColumnDefinition Width="Auto" />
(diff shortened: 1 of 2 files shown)


nmetulev:
Clicking stop should hide the button too, no?

brunosonnino:
It hides, but just after the first token. While it haven't received the token, it cannot be canceled

azchohfi:
> It hides, but just after the first token. While it haven't received the token, it cannot be canceled

Not even if you do:
```cs
await describeTask.AsTask(_cts.Token);
```
?

brunosonnino:
Sort of - it cancels the task, but the task continues in the background, so it's not canceled (although we think it is)

azchohfi:
But does it immediately throw an exception so we can instantaneously update the UI? If so, we should be good.
They might need to add cancellation support on their end, yes, but we are doing the right thing on our end.

azchohfi (approved):

PR #246, opened 2025-02-26
Moved to ORTGenAI's IChatClient.

azchohfi:


azchohfi:
Waiting for new version of ortgenai that will have this.

AIDevGallery/ProjectGenerator/Generator.cs
```diff
@@ -296,8 +296,7 @@ static void AddPackageReference(ProjectItemGroupElement itemGroup, string packag
                     packageReferenceItem.Condition = "$(Platform) == 'x64'";
                 }
                 else if (packageName == "Microsoft.ML.OnnxRuntime.Qnn" ||
-                            packageName == "Microsoft.ML.OnnxRuntimeGenAI" ||
-                            packageName == "Microsoft.ML.OnnxRuntimeGenAI.Managed")
+                            packageName == "Microsoft.ML.OnnxRuntimeGenAI")
```
nmetulev: We should move to Microsoft.ML.OnnxRuntimeGenAI.QNN here now that it's available (should simplify the csproj also)

azchohfi: Oh, that is interesting. I didn't see that before. Cool!

azchohfi: Should be solved now. Thanks for this, I love how much simpler the projects are now 💖

azchohfi:
Build should pass, but arm64 is not working yet, since rc1 doesn't have arm64 binaries on Microsoft.ML.OnnxRuntimeGenAI.

azchohfi:
Fixed the arm64 issue.

azchohfi:
rc2 has arm64 assemblies, so this should build. Still, it is not working as we need the SK fix. We should merge this only after #292, which will have the SK update.
https://github.com/microsoft/semantic-kernel/pull/10989

azchohfi:
@nmetulev we should be unblocked, but we should probably wait until stable 0.7.0 to merge this.

nmetulev (approved):

PR #247, opened 2025-02-26
Updated test UI, adding open build folder and open logs buttons.

azchohfi:
image


nmetulev:
Could you add a bit of docs about how to run these test manually? Maybe in our docs folder?

azchohfi:
These are just simple MSTest tests. They show up in test explorer in VS, and you simply run them.

PR #263, opened 2025-03-03
Users can add local GenAI models. also remembers execution provider in MRU

nmetulev:


AIDevGallery.Utils/ModelUrl.cs
```diff
@@ -252,6 +252,10 @@ public static string GetFullUrl(string url)
         {
             return url;
         }
+        else if (url.StartsWith("local", StringComparison.InvariantCultureIgnoreCase))
```
tarekgh: >StringComparison.InvariantCultureIgnoreCase [](http://example.com/codeflow?start=41&length=43)

nit: I think you need to use `OrdinalIgnoreCase` better. It is faster and avoid running into linguistic cases in the comparison. I am seeing InvariantCulture used in many places which doing the operation as ordinal comparison make more sense.

nmetulev: Thank you @tarekgh - I updated all instances

AIDevGallery/Helpers/ModelDetailsHelper.cs
```diff
@@ -155,11 +155,20 @@ public static Visibility ShowWhenOllama(ModelDetails modelDetails)
...
             || modelDetails.HardwareAccelerators.Contains(HardwareAccelerator.DML)
             || modelDetails.HardwareAccelerators.Contains(HardwareAccelerator.QNN)
             ? Visibility.Visible : Visibility.Collapsed;
     }
+
+    public static Visibility ShowWhenDownloadedModel(ModelDetails modelDetails)
+    {
+        return (modelDetails.HardwareAccelerators.Contains(HardwareAccelerator.CPU)
+            || modelDetails.HardwareAccelerators.Contains(HardwareAccelerator.DML)
+            || modelDetails.HardwareAccelerators.Contains(HardwareAccelerator.QNN))
```
azchohfi: Refactor to use the logic from ShowWhenOnnxModel.

AIDevGallery/Pages/Models/AddModelPage.xaml.cs
```diff
@@ -331,6 +333,150 @@ private void SearchTextBox_Loaded(object sender, RoutedEventArgs e)
     {
         this.Focus(FocusState.Programmatic);
     }
+
+    private async void AddLocalClicked(object sender, RoutedEventArgs e)
```
azchohfi, line 351: A large portion of the logic in this method is the same, or an upgrade (QNN), of the logic in the SearchModels method in this class. This should be abstracted so the improvements here are also shared with HF models, and we keep the logic in one place. I'm thinking in a method that accepts a GenAIConfig and returns a ModelDetails. Differences can still exist on the object, but there is a lot of shared code here, so they should be in the same method.

AIDevGallery/Pages/Models/ModelPage.xaml.cs
```diff
@@ -292,4 +301,14 @@ private void SampleList_ItemInvoked(ItemsView sender, ItemsViewItemInvokedEventA
             App.MainWindow.Navigate("Samples", new SampleNavigationArgs(sample, availableModel));
         }
     }
+
+    public static Uri GetSafeUrl(string? url)
+    {
+        if (string.IsNullOrWhiteSpace(url))
+        {
+            return new Uri("https://aka.ms/ai-dev-gallery-repo");
```
azchohfi, line 309: Wouldn't it be better to move this to the SG? Make the property non-nullable and just generate it there. No converter, no null.

AIDevGallery/Samples/Open Source Models/Image Models/ESRGAN/SuperResolution.xaml.cs
```diff
@@ -56,7 +59,10 @@ protected override async Task LoadModelAsync(SampleNavigationParameters samplePa
     {
         var hardwareAccelerator = sampleParams.HardwareAccelerator;
         await InitModel(sampleParams.ModelPath, hardwareAccelerator);
+
         sampleParams.NotifyCompletion();
+
+        await EnhanceImage(Windows.ApplicationModel.Package.Current.InstalledLocation.Path + "\\Assets\\Enhance.png");
```
azchohfi: Use Path.Join.

AIDevGallery/Utils/AppUtils.cs
```diff
@@ -171,14 +171,19 @@ public static string GetHardwareAcceleratorDescription(HardwareAccelerator hardw
 
     public static string GetModelSourceNameFromUrl(string url)
```
azchohfi: This method's meaning is changed, so I would rename the method.

azchohfi (changes requested):

AIDevGallery/Helpers/ModelDetailsHelper.cs
nmetulev: done

azchohfi (approved):

PR #275, opened 2025-03-05
Add ShowExceptionmethod to BaseSamplePage

### LINKED ISSUE #175: [BUG] [WCR] APIs can get into a bad state that requires machine reboot
**Describe the bug**
When APIs get into a bad state, capture the exception and let the user know that there is an issue with these experimental APIs and a machine reboot should fix it.

### LINKED ISSUE #176: [SAMPLE] SampleContainer should have an exception UI
Whenever a sample would have an exception, the sample could trigger a `ShowException(Exception ex, string? optionalMessage') in `SampleContainer` to show details to the user in a consistent way and with a potential "Reload sample" button

brunosonnino:
Fixes #175
Fixes #176


AIDevGallery/ProjectGenerator/Template/MainWindow.xaml.cs
```diff
@@ -17,4 +20,30 @@ internal void ModelLoaded()
     {
         ProgressRingGrid.Visibility = Visibility.Collapsed;
     }
+
+    internal async void ShowException(Exception ex, string? optionalMessage = null)
+    {
+        var msg = optionalMessage ?? ex switch
+        {
+            COMException
```
nmetulev, line 30: IMO, we should have this code in the sample itself as this method can be generalized for all samples, not just for WCR samples. The ShowException method should be very general

AIDevGallery/Samples/BaseSamplePage.cs
```diff
@@ -15,13 +18,29 @@ internal partial class BaseSamplePage : Page
     protected override async void OnNavigatedTo(NavigationEventArgs e)
     {
         base.OnNavigatedTo(e);
+        App.Current.UnhandledException += Current_UnhandledException;
```
nmetulev: Wouldn't this handle all app exception and not just the sample exception?

AIDevGallery/Samples/BaseSamplePage.cs
```diff
@@ -42,4 +61,38 @@ internal void SendSampleInteractedEvent(string? customInfo = null)
...
+    internal async void ShowException(Exception ex, string? optionalMessage = null)
+    {
+        var msg = optionalMessage ?? ex switch
+        {
+            COMException
+                when ex.Message.Contains("the rpc server is unavailable", StringComparison.CurrentCultureIgnoreCase) =>
+                    "The WCL is in an unstable state.\nRebooting the machine will restart the WCL.",
+            _ => $"Error:\n{ex.Message}{(optionalMessage != null ? "\n" + optionalMessage : string.Empty)}"
+        };
+        ContentDialog exceptionDialog = new()
```
nmetulev, line 104: We should have a copy exception details button here

nmetulev (changes requested):
This is great - we should make this more generic so it can be used by all samples.

AIDevGallery/ProjectGenerator/Template/MainWindow.xaml.cs
azchohfi: It could live in a simple static class (in a SharedCode file). The file would use the XamlRoot from the static App.Window?.Content.XamlRoot (probably) and same for RootFrame.

AIDevGallery/Samples/WCRAPIs/ImageDescription.xaml.cs
```diff
@@ -201,13 +201,13 @@ private async void DescribeImage(SoftwareBitmap bitmap)
...
 
-        Loader.Visibility = Visibility.Collapsed;
-        StopBtn.Visibility = Visibility.Collapsed;
-        _cts?.Dispose();
-        _cts = null;
+            Loader.Visibility = Visibility.Collapsed;
+            StopBtn.Visibility = Visibility.Collapsed;
+            _cts?.Dispose();
+            _cts = null;
+        }
```
azchohfi: Shouldn't this be in the finally block, or outside the catch?

AIDevGallery/Samples/SharedCode/IChatClient/PhiSilicaClient.cs
```diff
@@ -60,6 +61,10 @@ private static ChatOptions GetDefaultChatOptions()
         {
             await phiSilicaClient.InitializeAsync(cancellationToken);
         }
+        catch (COMException ex) when (ex.Message.Contains("the rpc server is unavailable", System.StringComparison.CurrentCultureIgnoreCase))
```
azchohfi: If you are handling this here and throwing, this raises the question of if we should handle any exception here at all. I mean, why return null and not throw always, if there is an exception that is being handled elsewhere?

AIDevGallery/Samples/BaseSamplePage.cs
brunosonnino: Yes, it would handle all app exceptions while the sample is open. When the sample closes, the handler is removed

AIDevGallery/Samples/SharedCode/IChatClient/PhiSilicaClient.cs
brunosonnino: This is something to question. The code was returning null and I left it as it was. But I agree that we should throw and let the sample deal with it

AIDevGallery/ProjectGenerator/Template/MainWindow.xaml.cs
azchohfi: Lets not do this, as it would make things more complex.

AIDevGallery/Samples/WCRAPIs/ImageDescription.xaml.cs
brunosonnino: Yes, I will change that

AIDevGallery/Samples/BaseSamplePage.cs
azchohfi: This is added now.

azchohfi (approved):

AIDevGallery/Samples/SharedCode/IChatClient/PhiSilicaClient.cs
azchohfi: Lets not deal with this now.

PR #288, opened 2025-03-10
[Sample] Background detector sample

Jaylyn-Barbee:
Need to make sure we note that this sample only works well for **Portrait style images** and not general photos.


AIDevGallery/Samples/scenarios.json
```diff
@@ -156,6 +156,11 @@
         "Description": "Remove image backgrounds with an image segmentation model. Upload an image and select what objects in the image you want to include by clicking on locations in the image. Select remove background to get the extracted objects without the background.",
         "Id": "background-remover"
       },
+      "BackgroundDetector": {
+        "Name": "Detect Background",
+        "Description": "Detect image foreground vs background.",
```
nmetulev:
```suggestion
        "Description": "Detect image foreground vs background. Upload a portrait image and the model will separate the foreground from the background.",
```

nmetulev (approved):

PR #289, opened 2025-03-11
Adding Ollama support

nmetulev:


AIDevGallery/Helpers/ModelDetailsHelper.cs
```diff
@@ -130,4 +131,35 @@ static List<ModelDetails> GetSamplesForModelType(ModelType initialModelType)
...
+    {
+        return modelDetails.HardwareAccelerators.Contains(HardwareAccelerator.WCRAPI) ||
+               modelDetails.HardwareAccelerators.Contains(HardwareAccelerator.OLLAMA) ||
+               modelDetails.Size == 0;
+    }
+
+    public static bool IsApi(this ExpandedModelDetails modelDetails)
+    {
+        return modelDetails.HardwareAccelerator == HardwareAccelerator.WCRAPI ||
+            modelDetails.HardwareAccelerator == HardwareAccelerator.OLLAMA;
```
azchohfi, line 145: Would be best to have a shared 3rd method that accepts a list of HardwareAccelerators, and does the `.Contains(...)`.

AIDevGallery/Pages/Scenarios/ScenarioPage.xaml.cs
```diff
@@ -137,6 +160,15 @@ private async void ModelSelectionControl_SelectedModelChanged(object sender, Mod
         {
             foreach (var s in samples)
             {
+                if (selectedModelDetails.HardwareAccelerators.Contains(HardwareAccelerator.OLLAMA))
```
azchohfi, line 163: What is this comparing? Only the hardware accelerator? This seems wrong, but I'm not sure.

AIDevGallery/ProjectGenerator/Generator.cs
```diff
@@ -134,6 +134,10 @@ internal async Task<string> GenerateAsync(Sample sample, Dictionary<ModelType, E
                     modelIds.Add("UserAdded");
                 }
             }
+            else
+            {
+                modelIds.Add(modelInfo.Id);
```
azchohfi, line 139: Shouldn't we check if it is an actual Ollama model here? I would not assume they are.

AIDevGallery/Utils/Ollama.cs
```diff
@@ -0,0 +1,76 @@
...
+                        var nameTag = tokens[0].Split(':');
+
+                        models.Add(new OllamaModel(nameTag[0], nameTag[1], tokens[1], tokens[2], tokens[3]));
+                    }
+                }
+
+                return models;
+            }
+        }
+        catch
```
azchohfi: Should we handle the presence of ollama or not? This seems too generic. We can easily check the Path:
https://github.com/microsoft/msstore-cli/blob/3801a81b9a955b6af20e1ef695f73c952c57de58/MSStore.CLI/Services/ExternalCommandExecutor.cs#L103-L123
Also, should we have this as a setting if it is not on the Path?

AIDevGallery/Helpers/ModelDetailsHelper.cs
nmetulev: We could add it we need it. IsApi feels generic enough for now.

AIDevGallery/Pages/Scenarios/ScenarioPage.xaml.cs
nmetulev: The modelDetails for Ollama models are not in the dictionary of models so the id will never be found with GetModelDetails if the selected model is Ollama, so this just checks if the selected model is Ollama and if the sample calls for LanguageModel

AIDevGallery/ProjectGenerator/Generator.cs
nmetulev: I have it more as a catch all if all other elses fail. Not setting modelIds here will throw an exception bellow when we try to access modelIds. But you know this code better, so not sure if there is a better solution

AIDevGallery/Utils/Ollama.cs
nmetulev: I thought about it but I thought to keep it simple and just let the check for ollama fail. Good idea about caching that check, I'll add that

AIDevGallery/Utils/Ollama.cs
```diff
@@ -0,0 +1,89 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading.Tasks;
+
+namespace AIDevGallery.Utils;
+
+internal record OllamaModel(string Name, string Tag, string Id, string Size, string Modified);
+
+internal class Ollama
```
azchohfi: Change class name to something more meaningful.

zateutsch (approved):
LGTM - the only thing I wonder about is where to indicate that this is available in the app, but I think that can be added later on

AIDevGallery/Utils/Ollama.cs
nmetulev: Done

AIDevGallery/Utils/OllamaHelper.cs
```diff
@@ -0,0 +1,89 @@
...
+                p.StartInfo.RedirectStandardError = true;
+                p.StartInfo.UseShellExecute = false;
+                p.StartInfo.CreateNoWindow = true;
+
+                p.Start();
+                p.WaitForExit();
+
+                var lines = new List<string>();
+
+                Task.WhenAny(
```
azchohfi, line 39: Why is all of this needed? If the process has closed already, can't you just read the output?

PR #294, opened 2025-03-13
Fix API deeplinking

niels9001:
This feels a bit like a band-aid fix.. we should probably refactor the whole navigation and deeplinking code.

For now, this PR fixes the linking to an API page since that was broken + a UI fix for the copylink button positioning

Steps to repro:
- Navigate to a WCR API page like OCR
- Click the copy button
- Paste the link in the run box
- AI Dev Gallery should launch and navigate to the OCR page

Other deeplinks to test:

_aidevgallery://apis/4bcc0137-0e9a-4eda-8096-b235fcb0e98b_

_aidevgallery://scenarios/generate-text_

_aidevgallery://models/phi4mini_

### DIFF
--- a/AIDevGallery/Helpers/ActivationHelper.cs
+++ b/AIDevGallery/Helpers/ActivationHelper.cs
@@ -7,7 +7,6 @@ using AIDevGallery.Telemetry.Events;
 using Microsoft.Windows.AppLifecycle;
 using System.Linq;
 using Windows.ApplicationModel.Activation;
-using Windows.Data.Xml.Dom;
 
 namespace AIDevGallery.Helpers;
 
--- a/AIDevGallery/Helpers/ActivationHelper.cs
+++ b/AIDevGallery/Helpers/ActivationHelper.cs
@@ -7,6 +7,7 @@ using AIDevGallery.Telemetry.Events;
 using Microsoft.Windows.AppLifecycle;
 using System.Linq;
 using Windows.ApplicationModel.Activation;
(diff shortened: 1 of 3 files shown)


AIDevGallery/Helpers/ActivationHelper.cs
```diff
@@ -7,6 +7,7 @@
 using Microsoft.Windows.AppLifecycle;
 using System.Linq;
 using Windows.ApplicationModel.Activation;
+using Windows.Data.Xml.Dom;
```
azchohfi:
```suggestion
```

AIDevGallery/MainWindow.xaml.cs
```diff
@@ -12,6 +14,7 @@
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using Windows.Data.Xml.Dom;
```
azchohfi:
```suggestion
```

AIDevGallery/MainWindow.xaml.cs
```diff
@@ -2,8 +2,10 @@
 // Licensed under the MIT License.
 
 using AIDevGallery.Controls;
+using AIDevGallery.Helpers;
 using AIDevGallery.Models;
 using AIDevGallery.Pages;
+using AIDevGallery.Samples;
```
azchohfi:
```suggestion
```

azchohfi (changes requested):

azchohfi (approved):

PR #295, opened 2025-03-13
Replaced markdown control with WCT Labs version 

nmetulev:


niels9001:
Switching themes while running the app results in colors not being updated. Not a big deal though, bit of an edge case IMO

image

niels9001:
image

Linespacing is more than in the previous version

niels9001 (approved):
Two regressions, but nothing major so I'm fine with merging this

- Linespacing for text lines
- Switching themes does not update the colors

nmetulev:
I fixed switching the themes. I'll merge now and we can revisit the line numbers another time

PR #296, opened 2025-03-15
Replaced code viewer in sample view

nmetulev:
Removed pivot and replaced with tabs, added line numbers

image


niels9001 (approved):
Works well.. some nice polish 🚀

PR #305, opened 2025-03-19
#251: Added Support for AI Toolkit deeplinks

### LINKED ISSUE #251: [FEATURE] AI toolkit deep linking
Link to AI toolkit for models that are available for fine tuning there.

Eventually, link directly to model fine tuning view.

zateutsch:
fixes #251 

@niels9001 design pass would be great

image


AIDevGallery/Pages/Models/ModelPage.xaml.cs
```diff
@@ -155,6 +163,74 @@ private IEnumerable<ModelDetails> GetAllSampleDetails()
...
+        AIToolkitDropdown.Visibility = isAiToolkitActionAvailable ? Visibility.Visible : Visibility.Collapsed;
+    }
+
+    private void ToolkitActionFlyoutItem_Click(object sender, RoutedEventArgs e)
+    {
+        if(sender is MenuFlyoutItem actionFlyoutItem)
+        {
+            (AIToolkitAction action, ModelDetails modelDetails) = ((AIToolkitAction, ModelDetails))actionFlyoutItem.Tag;
+
+            AIToolkitActionClickedEvent.Log(AIToolkitHelper.AIToolkitActionInfos[action].QueryName, modelDetails.Name);
```
nmetulev: Would be good to also capture whether the deep link worked or we had to fallback to the ai toolkit docs. Can you add a property (bool) and call this below?

nmetulev (commented):
Amazing work!

AIDevGallery.SourceGenerator/Models/Model.cs
```diff
@@ -24,4 +24,7 @@ internal class Model
     [JsonConverter(typeof(SingleOrListOfStringConverter))]
     [JsonPropertyName("FileFilter")]
     public List<string>? FileFilters { get; init; }
+    public List<AIToolkitAction>? AIToolkitActions { get; set; }
```
azchohfi: Init for these properties, so they can't be changed.

AIDevGallery/Utils/AIToolkitHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+        { AIToolkitAction.BulkRun, new ToolkitActionInfo() { DisplayName = "Bulk Run", QueryName = "open_bulk_run" } },
+        { AIToolkitAction.Playground, new ToolkitActionInfo() { DisplayName = "Playground", QueryName = "open_playground" } }
+    };
+
+    public static Dictionary<AIToolkitAction, ToolkitActionInfo> AIToolkitActionInfos
+    {
+        get { return aiToolkitActionInfos; }
+    }
+
+    public static string CreateAiToolkitDeeplink(AIToolkitAction action, ModelDetails modelDetails)
```
azchohfi: Do these maybe make the methods on this class as extension methods? Would make using them a little bit nicer. Just a suggestion.

AIDevGallery/Utils/AIToolkitHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            deeplink = deeplink + $"{actionInfo.QueryName}?model_id={modelId}&track_from=AIDevGallery";
+        }
+
+        return deeplink;
+    }
+
+    public static bool ValidateForFineTuning(ModelDetails modelDetails)
+    {
+        return modelDetails.AIToolkitActions!.Contains(AIToolkitAction.FineTuning) && !string.IsNullOrEmpty(modelDetails.AIToolkitFinetuningId);
```
azchohfi: Should this method throw if called for a model without aitoolkitactions? Probably not so I would check for null, not "!." .

AIDevGallery/Utils/AIToolkitHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public static bool ValidateForGeneralToolkit(ModelDetails modelDetails)
+    {
+        return modelDetails.AIToolkitActions!.Where(action => action != AIToolkitAction.FineTuning).ToList().Count > 0 && !string.IsNullOrEmpty(modelDetails.AIToolkitId);
```
azchohfi: Same here.

AIDevGallery/Utils/AIToolkitHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+        else
+        {
+            return ValidateForGeneralToolkit(modelDetails);
+        }
+    }
+}
+
+internal class ToolkitActionInfo
+{
+    public required string DisplayName { get; set; }
```
azchohfi: Init here too.

AIDevGallery.SourceGenerator/Models/Model.cs
```diff
@@ -24,4 +24,7 @@ internal class Model
...
+    public string? AIToolkitId { get; set; }
+    public string? AIToolkitFinetuningId { get; set; }
```
azchohfi: Is this property being used at all? Or will it be used in the future? Seems like you would use one or the other only, no? So maybe it can be simplified to a single property.

azchohfi (changes requested):

azchohfi:
Amazing, btw!

niels9001:
@zateutsch Design pass done.. looking good to me!

image

AIDevGallery.SourceGenerator/Models/Model.cs
zateutsch: It will be used in the future once they add the finetuning deeplink - thought I'd just get ahead of it but can remove if necessary.

azchohfi: If we don't know when that is coming, I would remove it, but if it is a follow up PR that is coming soon, I'm ok with keeping it.

AIDevGallery/Utils/AIToolkitHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        if(aiToolkitActionInfos.TryGetValue(action, out actionInfo) && !string.IsNullOrEmpty(modelId))
+        {
+            deeplink = deeplink + $"{actionInfo.QueryName}?model_id={modelId}&track_from=AIDevGallery";
+        }
+
+        return deeplink;
+    }
+
+    public static bool ValidateForFineTuning(this ModelDetails modelDetails)
```
azchohfi, line 39: Since these are extension methods, I would make sure that all places calling them are using modelDetails.ValidateForFineTuning(). Same for all methods.

AIDevGallery/Utils/AIToolkitHelper.cs
azchohfi: Now that you changed the other methods to extension methods, this one is missing 😅

azchohfi (changes requested):

nmetulev (approved):

AIDevGallery/Pages/Models/ModelPage.xaml.cs
```diff
@@ -155,6 +164,82 @@ private IEnumerable<ModelDetails> GetAllSampleDetails()
...
     }
 
+    private void BuildAIToolkitButton()
+    {
+        bool isAiToolkitActionAvailable = false;
+        Dictionary<AIToolkitAction, MenuFlyoutSubItem> actionSubmenus = new();
+
+        foreach(ModelDetails modelDetails in models)
+        {
+            foreach(AIToolkitAction action in modelDetails.AIToolkitActions!)
```
azchohfi: This list can be null, so check for null. ! is very risky.

azchohfi (changes requested):

azchohfi (approved):

PR #308, opened 2025-03-20
Removed NumSharp.

azchohfi:
There is no need to use this library, as it is not native aot compatible, and we use only one method from it (Linspace, which is very simple to replace).


nmetulev:
@Jaylyn-Barbee, can you review this? feel free to merge if it looks good

PR #311, opened 2025-03-25
Adding error handling for current state of Intel C+PCs

Jaylyn-Barbee:
At the time of this PR only the OCR model works fully on Intel Copilot+ PCs. 

You are able to download both the ImageScaler and ImageSegmentation models, via `.isAvailable` or `.MakeAvailable` however, they fail at the `.CreateAsync` step. 

Both Phi Silica and Image describer fail at the `.isAvailable` step which makes them "undetectable" when on an Intel Copilot+ PC.

This enables the Intel NPU to be found and adds error handling to handle the above errors. Unfortunately, for the two models that fail at `.isAvailable` we are unable to tell whether that means we are not on a C+PC or if we are and they just aren't supported yet, so it shows the "You need a C+PC to use this sample" (paraphrasing) message. The only work around to fix this message would be assuming having an NPU implies C+PC but this is not necessarily always true.


nmetulev (approved):

Jaylyn-Barbee:
Need to cross check this PR on QC to ensure fixing one platform didn't cause problems on the other.

Jaylyn-Barbee (commented):
Everything still seems to be working as expected on QC C+PCs

PR #324, opened 2025-04-01
#321: Single model case for AI Toolkit Button

### LINKED ISSUE #321: [ENHANCEMENT] Don't require double click for AI toolkit button with single model
If there is only one model available to open in AI toolkit, we should just show the toolkit options without expanding the model option.

zateutsch:
fixes #321


nmetulev:
Opening a model in the playground doesn't load that model and I have to select it again, even if it was already downloaded in the toolkit. Is this a bug with the toolkit or the gallery?

nmetulev (approved):

PR #325, opened 2025-04-04
Added OpenAI Models.

azchohfi:


nmetulev:
This is amazing. Couple things
* Loading a sample now takes a bit longer and shows an emptish page before actually loading - should we add a loading screen or move the logic to the model picker? /cc @niels9001
* We need to include the right usings in the sample.xaml.cs depending on the IChatCLient implementation (and probably for other providers later, ie Azure)

azchohfi:
We could have an initialize call instead, that happens elsewhere, and then the returns are always sync, returning from a state they keep internally.

nmetulev (approved):

PR #326, opened 2025-04-04
[DRAFT] Ollama function calling

zateutsch:
Need help on model selection logic @azchohfi


AIDevGallery/ExternalModelUtils/IExternalModelProvider.cs
```diff
@@ -17,9 +17,10 @@ internal interface IExternalModelProvider
     string DarkIcon { get; }
     HardwareAccelerator ModelHardwareAccelerator { get; }
     List<string> NugetPackageReferences { get; }
+    private static List<string>? ToolCallingModelNames { get; }
```
azchohfi: This is an interface, so **if** this property should be implemented by all implementors, it should not be private (or static). This actually doesn't work. Also, it doesn't feel like this is the case, so I would not add this here, only on the OllamaModelProvider. It is an implementation detail.

AIDevGallery/Samples/Open Source Models/Language Models/ToolCalling.xaml.cs
```diff
@@ -0,0 +1,235 @@
...
+        };
+
+        this.InitializeComponent();
+    }
+
+    protected override async Task LoadModelAsync(SampleNavigationParameters sampleParams)
+    {
+        try
+        {
+            await Task.Run(() => Thread.Sleep(1));
```
azchohfi: Why is this needed?

AIDevGallery/Samples/Definitions/Models/languagemodels.modelgroup.json
```diff
@@ -199,5 +199,12 @@
         "ReadmeUrl": "https://huggingface.co/microsoft/mistral-7b-instruct-v0.2-ONNX/blob/main/README.md"
       }
     }
+  },
+  "ToolCallingLanguageModels": {
```
azchohfi: I don't think this will work on the long run... Isn't a "tool calling language model", just a language model that can run tools? This would just be a property of all LanguageModels, so each definition would need to declare if they can run tools or not. We don't have a mechanism for this, so I'm not sure on the solution.
To understand concern, I think it might be easier for you to just search for "ModelType.LanguageModels". Our existing logic relies on all language models being a child of LanguageModels. All the logic for ORTGenAI, project generation, etc, relies on this.

azchohfi (changes requested):

zateutsch:
@azchohfi I addressed your feedback - I am checking for the scenario type now in conjunction with the the ModelType being language model. This will need to reimagined in the future if we ever have a local model that can call function, but the work should be minimal.

@nmetulev Added you for review on this, but there's still a couple thing that need to be done (I need to get a full list of validated tool calling ollama models and actually polish up the sample), but thought you should see the update I made for the visual state if they select this sample without having ollama or one of the required models.

Issue #277, opened 2025-03-05
[BUG] Problems handling the Enter key in a prompt textbox in languages other than English


tokawa-ms:
**Describe the bug**
This is a problem report about a one-line text box for prompt input, such as on the "Generate Text" screen.

When entering a language other than English (such as Japanese) in this text box, the generation process is called because it is mistakenly judged that prompt input is completed while the character string is being entered.

**To Reproduce**
Requirements :
1. Japanese or other IMEs available in Windows environments

1. Go to 'Generate Text'
2. Enable IME
3. Type some charactors
4. Confirm the conversion candidate with the Enter key
5. Even though I'm in the middle of a sentence, GenerateText is called.

**Expected behavior**
- The ENTER keystroke to confirm the conversion of an IME does not invoke the generation process.
- When you complete the input of a sentence, a single Enter key is used to invoke the generation process.

**Please complete the following information:**
 - App Version : Version 0.3.6.0 from Microsoft Store and Latest Build from GitHub main branch.

**Additional context**
The reason for this is that in languages other than English, you press the Enter key to confirm the IME conversion.
The current logic is to call the generation process (e.g. GenerateText() method) whenever there is an Enter key input after a character is entered in TextBox.
However, if we are considering IME conversion, we should add an implementation that ignores the IME conversion confirmation Enter.

I have an idea to solve this problem.

Because the KeyDown event does not occur during IME input, it can be determined that the IME is being entered when the Enter key has not been entered in the previous KeyDown event. By adding this condition to the KeyUp event handler, it is possible to ignore Enter when the IME conversion is confirmed and call the generation process only with Enter when the string input is completed.

For example, I have been able to confirm that if I rewrite the event handler of the TextBox in Generate.xaml.cs as follows, it behaves as expected:

```csharp
    private bool isImeActive = true;
    private void TextBox_KeyUp(object sender, KeyRoutedEventArgs e)
    {
        if (e.Key == Windows.System.VirtualKey.Enter && sender is TextBox && isImeActive == false)
        {
            if (InputTextBox.Text.Length > 0)
            {
                GenerateText(InputTextBox.Text);
            }
        }

        isImeActive = true;
    }

    private void TextBox_KeyDown(object sender, KeyRoutedEventArgs e)
    {
        if (e.Key == Windows.System.VirtualKey.Enter)
        {
            isImeActive = false;
            e.Handled = true;
        }
    }
```

[...]

PR #341, opened 2025-04-22
Additional Fix for the issue #277 (IME Handling) in RAG Page.

tokawa-ms:
# Overview
In this PR, I added an implementation to each TextBox that handles the Enter keystroke, which means that the IME conversion is confirmed for RAG Page. (Same as PR #301 )
# Description
This code is proposed as a workaround for the following issues:
https://github.com/microsoft/ai-dev-gallery/issues/277

When entering a language that uses IME conversion (for example, Japanese) in the text box and confirming IME conversion, the input string is incorrectly judged to be text input confirmed and the input string is sent in the middle.

So, in this PR, I've added an implementation to each TextBox that handles the Enter keystroke, which means that the IME conversion is confirmed.

It works the same the before version in English, and I've verified that the IME works as expected in Japanese, but there may still be glitches in other languages. I think it would be better if you could get reviews from people in other languages.

# Logic
This proposed fix takes advantage of WPF's behavior that the PreviewKeyDown event is not raised for a TextBox when the IME is enabled (I'm not sure if this is documented) to determine whether an IME is enabled.


nmetulev (approved):

nmetulev:
Thank you @tokawa-ms

PR #357, opened 2025-04-28
New model picker

nmetulev:


azchohfi (commented):
A few feedback items:

1. The WCR APIs screen feels too empty. If there is only one item, we should have a better image there, saying what it is, etc.
image
2. Separate, but related, If Copilot+PC is required for a specific sample, it should show only a small header or something like that, and the code tab should be automatically expanded (its a little bit of a waste of space):
image
We should make the code tab expanded by default if you can't run the sample.
3. There is something about this new UX that feels weird. Maybe its because there are not many fluid animations/transitions, and it just blinks on the screen when I move to a new sample. Also, this:
image
Spacing seems off.
4. The flow for multiple models is also weird:
image
Why can't I hit save? I know why, but its not clear. Maybe make that Save button enabled, but change the text to "Next", which will select the next model?

In general, this is much better than the experience we have now, but I did want to call these out.
I believe that for samples with a single model it should hide the right pane, and for multiple models it needs to be clearer in what it is that the user is expected to do next.

niels9001:
@azchohfi I've pushed a few small UX tweaks (inc. some animations) that hopefully address some of your topics.

Added a title to the model selection ("Selected models for this sample") here as well.
image

niels9001:
I wonder if we need some sort of title for the multi-models that would make this it a bit clearer.. E.g. 

"Selected language model"
"Selected classification model"

We could even have a little i tooltip to explain what each model does

niels9001 (approved):

niels9001 (approved):

PR #366, opened 2025-05-09
UX tweaks to scenariopage

niels9001:
Some visual tweaks, and moving the sample selector dropdown to the bottom of the page.

image


nmetulev (commented):
We should move the bottom bar to the sample container so the warning shows up in the api page as well.

Narrow view things overlap, should we have the sample dropdown move to just an icon at some point?

fivemserverhostings:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

nmetulev (approved):

gregwoo-microsoft (approved):

Issue #382, opened 2025-05-24
[BUG] Select_Text_From_ImageSample crash bug


AngleHony:
**Describe the bug**
Select_Text_From_ImageSample,  TextRecognition.xaml.cs, When a file is selected, soems times  FontSize is zero, the app crash

**To Reproduce**
1. start Select_Text_From_ImageSample 
2. Click on 'from file '
3. select the image  

![Image](https://github.com/user-attachments/assets/728196ca-6fa0-41c7-9c4f-f34aee090a93)

4. app crash

**Expected behavior**
**Screenshots**

![Image](https://github.com/user-attachments/assets/24c3c3ad-516f-414c-a5f8-cf698f2a1517)

**Please complete the following information:**
 - App Version [e.g. 0.4.2.0 from the Microsoft Store]:

**Additional context**
FontSize = Math.Abs((int)line.BoundingBox.TopRight.Y - (int)line.BoundingBox.BottomRight.Y) * .85,

some case line.BoundingBox.TopRight.Y  == (int)line.BoundingBox.BottomRight.Y  then FontSize = 0;  

then app crash

nmetulev:
Thanks for reporting this - looks like we need to handle the case you described here. I can look at this sometime in the next two weeks, but feel free to submit a PR in the meantime

wang563681252:
This issue has been fixed in the latest version.

PR #385, opened 2025-05-26
fix when FontSize is set to zero,  the Select_Text_From_ImageSample app will crash

AngleHony:
fix https://github.com/microsoft/ai-dev-gallery/issues/382


AIDevGallery/Samples/WCRAPIs/TextRecognition.xaml.cs
```diff
@@ -207,7 +207,13 @@ private void RenderRecognizedText(RecognizedText recognizedText)
         List<string> lines = new List<string>();
 
         foreach (var line in recognizedText.Lines)
-        {
+        {
+            double fontSize;
```
brunosonnino, line 211: Making the assignment in line 211 and just comparing the variable with 0 in line 212 seems more clear to me, and this code should be added after line 217. Even if the text size is 0, we should add it to the recognized lines, no?

nmetulev:
Hey @AngleHony, are you able to address the PR feedback here?

PR #390, opened 2025-06-02
Text Intelligence APIs

brunosonnino:


nmetulev:
Awesome.

I wonder if we should combine these Phi Silica based samples somehow? 
- Option 1 is to add a Phi Silica parent and have all the pages as sub pages in the left navigation
- Option 2 is to include these samples in the Phi Silica page instead since the documentation is the same for most of them

We should also have the sample input prepopulated, especially for samples like text to table, to make it easier to demo

nmetulev:
Just moved the phi silica apis under their own category - @niels9001 and @brunosonnino, thoughts
image

niels9001:
> Just moved the phi silica apis under their own category - @niels9001 and @brunosonnino, thoughts
> image
> 

Ship it!

brunosonnino:
> Just moved the phi silica apis under their own category - @niels9001 and @brunosonnino, thoughts image

It's great. Just two minor nits: the TextRewriter and Text To Table have the same icon and SuperResolution and Object Erase should have their icons swapped

nmetulev:
Here are the icons updated 
image

nmetulev:
@brunosonnino, can you fix the warnings so your commit is the last one here so we can merge this?

nmetulev (approved):

PR #393, opened 2025-06-05
[StableDiffusion Sample][NvTensorRTRTX EP] Add free dimension overrides for SD1.4 components

keshavv27:
Initialization of Stable Diffusion pipeline in sample application was failing when run through NVTensorRTRTX Execution Provider. The initialization requires dynamic dimensions in model inputs to be set explicitly through free dimension overrides. Some model inputs had to be modified to support the overrides. Modified sample runs with CPU EP as well. 

This PR:
1. sets free dimension overrides corresponding to the dimension values used in sample
2. modifies timestep input for unet model from {timestep} with dim 1 to {timestep,timestep} with dim 2
3. removes tensor reordering for inputImagesTensor (input to safety checker model)

### DIFF
--- a/AIDevGallery/Samples/SharedCode/StableDiffusionCode/StableDiffusion.cs
+++ b/AIDevGallery/Samples/SharedCode/StableDiffusionCode/StableDiffusion.cs
@@ -88,7 +88,7 @@ internal class StableDiffusion : IDisposable
             sessionOptions.AddFreeDimensionOverrideByName("height", config.Height / 8);
             sessionOptions.AddFreeDimensionOverrideByName("width", config.Width / 8);
             sessionOptions.AddFreeDimensionOverrideByName("sequence", 77);
-            
+
             if (policy != null)
             {
                 sessionOptions.SetEpSelectionPolicy(policy.Value);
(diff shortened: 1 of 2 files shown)


nmetulev:
LGTM - thank you

AIDevGallery/Samples/SharedCode/StableDiffusionCode/TextProcessing.cs
```diff
@@ -14,6 +14,15 @@ namespace AIDevGallery.Samples.SharedCode.StableDiffusionCode;
...
 {
+    private readonly StableDiffusionConfig config = new()
+    {
+        // Number of denoising steps
+        NumInferenceSteps = 15,
+
+        // Scale for classifier-free guidance
+        GuidanceScale = 7.5
+    };
+
```
ishwar-raut1, line 25: This could have been reused from the stableDiffusion.  There was no need for the stable diffusion config. Ideally single config should be propagated  [StableDiffusion.cs].

keshavv27: Creating a separate PR for this, thanks for pointing this out!

Issue #388, opened 2025-05-27
[FEATURE] Clear/rewrite buttons in chat


alejandro5042:
Please consider adding a "Clear" button to the chat, so you can begin again; and/or a "Rewrite" button that puts the latest user message in the chat box again so you can try a variant.

nmetulev:
I think this is a great idea - we can add this to the backlog. Feel free to also submit a PR if this is something you want to do, just let me know.

haoliuu:
Feature change merged https://github.com/microsoft/ai-dev-gallery/pull/431. Should be included in next 0.5.1 release.

PR #431, opened 2025-08-28
Clear and Rewrite buttons in Chat Sample

haoliuu:
#### Issue
https://github.com/microsoft/ai-dev-gallery/issues/388

#### Description

Introduce "Clear Chat" and "Rewrite Last Message" buttons to the chat UI, allowing users to clear all messages or edit and resend the last user message.

image

 
**ToolTip of the buttons**
image
image

**Click the Rewrite Button**
image

### DIFF
--- a/AIDevGallery/Samples/Open Source Models/Language Models/Chat.xaml
+++ b/AIDevGallery/Samples/Open Source Models/Language Models/Chat.xaml
@@ -100,7 +100,20 @@
                 AcceptsReturn="True"
                 MaxHeight="148"
                 ScrollViewer.VerticalScrollBarVisibility="Auto"/>
-            <StackPanel Grid.Column="1" Orientation="Horizontal">
+            <StackPanel Grid.Column="1" Orientation="Horizontal" Spacing="8">
+                <Button x:Name="ClearBtn"
+                        AutomationProperties.Name="Clear Chat"
+                        Click="ClearBtn_Click"
+                        ToolTipService.ToolTip="Clear all messages">
+                    <FontIcon FontSize="16" Glyph="&#xE74D;" />
+                </Button>
+                <Button x:Name="RewriteBtn"
+                        AutomationProperties.Name="Rewrite Last Message"
+                        Click="RewriteBtn_Click"
+                        ToolTipService.ToolTip="Edit and resend last message"
+                        IsEnabled="False">
+                    <FontIcon FontSize="16" Glyph="&#xE70F;" />
+                </Button>
                 <Button x:Name="SendBtn"
                         AutomationProperties.Name="Send Message"
                         Click="SendBtn_Click"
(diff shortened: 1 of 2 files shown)


AIDevGallery/Samples/Open Source Models/Language Models/Chat.xaml
```diff
@@ -100,7 +100,20 @@
                 AcceptsReturn="True"
                 MaxHeight="148"
                 ScrollViewer.VerticalScrollBarVisibility="Auto"/>
-            <StackPanel Grid.Column="1" Orientation="Horizontal">
+            <StackPanel Grid.Column="1" Orientation="Horizontal" Spacing="8">
+                <Button x:Name="ClearBtn"
```
weiyuanyue: currently, both buttons are next to the input box, which makes them look like they are for clearing or editing the current input. But their actual function is more about the whole chat history. Would it make more sense to move 'Clear Chat' to the top-right corner of the chat box? And maybe 'Rewrite' could be closer to the last message itself?

AIDevGallery/Samples/Open Source Models/Language Models/Chat.xaml
```diff
@@ -100,7 +100,20 @@
...
+                        AutomationProperties.Name="Clear Chat"
+                        Click="ClearBtn_Click"
```
weiyuanyue: If the chat history is already empty, should we disable the 'Clear Chat' button? Clicking it when thereis nothing to clear might confuse users or make them think it it broken

AIDevGallery/Samples/Open Source Models/Language Models/Chat.xaml.cs
```diff
@@ -234,6 +246,32 @@ private void EnableInputBoxWithPlaceholder()
         InputBox.PlaceholderText = "Enter your prompt (Press Shift + Enter to insert a newline)";
     }
 
+    private void ClearChat()
+    {
+        Messages.Clear();
```
weiyuanyue, line 370: When the model is still streaming a text response, should we consider stopping the stream first if the user clicks 'Clear Chat'?

AIDevGallery/Samples/Open Source Models/Language Models/Chat.xaml
haoliuu: Make sense. I will redesign the button to adjust the position. Also I think we could add more content instead of a single icon to avoid confusion.

AIDevGallery/Samples/Open Source Models/Language Models/Chat.xaml
haoliuu: Yup. Adding one more check for the Message count can improve this.

AIDevGallery/Samples/Open Source Models/Language Models/Chat.xaml.cs
haoliuu: We can reuse the cancel response func of the stop button here

weiyuanyue (approved):

PR #448, opened 2025-09-24
Streamline LAF Token and Fix Release LAF Token issue for Phi

weiyuanyue:
### Summary
This change improves configuration ergonomics, UI reusability, and error transparency. It removes unused build constants, streamlines how the LAF token is provided via environment variables, encapsulates the WcrModelUnavailable UI into a parameterized control, and refactors LimitedAccessFeaturesHelper to support token injection and richer error handling. **Fixes a Release packaging-stage token retrieval failure that left Phisilica unusable by switching to MSBuild-based token injection via AssemblyMetadata, making the process deterministic, CI-safe, and decoupled from business logic files.**

### Key Changes
- Remove obsolete DefineConstants to reduce build complexity and ambiguity.
- Simplify and secure LAF token provisioning:
  - Support environment variable–based configuration (e.g., LAF_TOKEN).
  - Provide a single, testable entry point for token injection via metadata, avoiding fragile file-based replacements.
  - Clarify precedence and fallback behavior.
- Refactor LimitedAccessFeaturesHelper:
  - Accepts injected LAF token to decouple configuration.
  - Expands error taxonomy and maps errors to clearer, structured types to aid diagnostics and logging.
- Introduce a reusable WcrModelUnavailable control:
  - Encapsulates the previous StackPanel into a dedicated control.
  - Accepts parameters (e.g., title/description/status) for dynamic rendering instead of hardcoded XAML strings.
  - Displays simple error codes to surface unavailability reasons quickly.
image

### Rationale

1. Eliminating unused DefineConstants reduces maintenance overhead and prevents configuration drift.
2. Environment variable–based LAF token injection via MSBuild and AssemblyMetadata enables cleaner local and CI/CD configuration, improves testability, and reduces coupling. It also avoids runtime file modifications, improving security and build reliability.
3. A parameterized WcrModelUnavailable control centralizes presentation logic, avoids duplication, and makes error visibility first-class.
4. Enriching LimitedAccessFeaturesHelper’s error modeling improves debuggability and future-proofs handling for additional error scenarios.
5. Fixes a Release packaging-stage token retrieval failure that left Phisilica unusable by switching to MSBuild-based token injection via AssemblyMetadata, making the process deterministic, CI-safe, and decoupled from business logic files.


wang563681252:
I think the LafConstants.cs should not be deleted and then directly modify the LimitedAccessFeaturesHelper.cs file. This is because the change frequency of the LimitedAccessFeaturesHelper.cs file will be higher than that of LafConstants.cs. If LafConstants.cs is purely generated code, it will prevent developers from making modifications to this file during development, which could lead to pipelining build failures.

weiyuanyue:
@wang563681252 

Thanks for the advice! I agree that changing files like _LimitedAccessFeaturesHelper.cs_ often can cause problems for maintenance. To solve this and keep config separate from business logic, I now use MSBuild properties and AssemblyMetadata to pass values. These values are read using reflection when the app runs.
This way, we do not need to change any files directly. It also makes the app safer and helps avoid build issues. And it removes the need to replace placeholder strings, which could cause XSS risks.

haoliuu (approved):

haoliuu:
Discussed offline. Good to go and test on CI

PR #458, opened 2025-09-28
Fixed a memory leak issue related to event handlers

HotCakeX:
Event handlers in the HoverLight weren't being unsubscribed On Disconnected.

PR #460, opened 2025-09-28
Fixed home page carousel visually leaking to other pages after navigation

HotCakeX:
The carousel on the home page visually leaks to other pages after navigation, that means we can see the carousel on other pages briefly after clicking on the tiles in the home page carousel. This PR fixes this problem completely; however, you need to accept my previous PR first, otherwise this change won't work, and you will get an error after navigation: https://github.com/microsoft/ai-dev-gallery/pull/458/files

Please see the videos below for more details as they exactly show the issue this PR fixes:

<br>

### Before

<div align="center">

https://github.com/user-attachments/assets/8611d99b-cabf-45b3-865a-4622747f6dcb

</div>

---

### After

<div align="center">

https://github.com/user-attachments/assets/6b8fd385-d6c3-49b1-ba76-2c02b487a27a

</div>

<br>


weiyuanyue:
Appreciate the intent to ensure resources are released when navigating away. However, I don’t think we should add this change as-is:

- WinUI navigation already triggers Unloaded and tears down the visual tree. Forcibly clearing rootGrid.Children and setting this.Content = null can interfere with back navigation, page caching, and transition animations, and may cause blank pages or NREs on return.
- This assumes this.Content is a Grid, which isn’t guaranteed long-term and is fragile.
- Clearing mostRecentlyUsedItems is unnecessary; the page-level collection is repopulated on Loaded and will be GC’d when the page is disposed. If we need a fresh list per navigation, we can reset it in OnNavigatedTo/Loaded without tearing down the UI

HotCakeX:
Hi @weiyuanyue 
Thanks for reviewing my PR, I believe I've addressed all of the points you raised while fixing the visual bug shown in the video. Will you please do another review?

PR #467, opened 2025-10-13
[Fix]Exported samples: replace LAF Token/PublisherId with demo values

weiyuanyue:
- Background

  - In-app, demoToken and demoPublisherId are obtained at runtime via LimitedAccessFeaturesHelper, which is correct for AI Dev Gallery. However, exported standalone samples often lack local setup/required dependency, causing first-run failures.
  - Goal: make exported samples runnable without extra configuration, while keeping in-app behavior unchanged.

- What changed

  - Pass `forExport: true` when writing Sample.xaml.cs during export so changes only affect exported samples.
  - Apply global replacements (for all samples) in exported code:
    - `var demoToken = LimitedAccessFeaturesHelper.GetAiLanguageModelToken();` → `var demoToken = "_YOUR_LAF_TOKEN_";`
    - `var demoPublisherId = LimitedAccessFeaturesHelper.GetAiLanguageModelPublisherId();` → `var demoPublisherId = "_YOUR_PUBLISHER_ID_";`
    - before:
    image

    - after: 
    
image

  - Use regexes that match the method calls (with optional whitespace) and replace them with string literals for robustness.

- Why
  - Exported samples become truly out-of-the-box for demos/learning.
  - In-app behavior remains intact: runtime retrieval via LimitedAccessFeaturesHelper is unchanged.
 
- Compatibility and risk
  - Changes apply only to exported code paths; in-app paths are untouched.
  - Regexes match method invocations rather than whole lines, making them resilient to style differences.
  - Demo values are for non-production use; production apps should still obtain real LAF credentials.

### DIFF
--- a/docs/PublishingNewVersion.md
+++ b/docs/PublishingNewVersion.md
@@ -25,18 +25,21 @@ This document describes the steps to publish a new version of the AI Dev Gallery
 
     This will update the version number in the `version.json` file on the current branch (main), and create a commit with an updated version number. By default it will increment the patch version number, but you can find more information about nbgv's cli [here](https://github.com/dotnet/Nerdbank.GitVersioning/blob/main/docfx/docs/nbgv-cli.md).
 
-    It will also create a new branch with the name `rel/vX.Y.Z` where `X.Y.Z` is the `current` version number. This is the release branch, which will be automatically checked out after the command is executed.
+    It will also create a new branch with the name `rel/vX.Y.Z` where `X.Y.Z` is the `current` version number.
 
 4. Now you can push the changes to the remote repository by running the following command:
 
     ```console
+    git checkout rel/vX.Y.Z
     git push -u origin rel/vX.Y.Z
     ```
 
-5. To make sure eveybody is now working on the new release, checkout the main branch and push the changes to the remote repository:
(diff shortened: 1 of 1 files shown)


AIDevGallery/Helpers/SamplesHelper.cs
```diff
@@ -313,6 +319,13 @@ string GetValueOrNull(string? value, string ifNull)
             }
         }
 
+        // When exporting example projects, replace demoToken and demoPublisherId retrieval with a mock value in any sample file
+        if (forExport)
+        {
+            cleanCsSource = RegexLafTokenAssignment().Replace(cleanCsSource, "\"_YOUR_LAF_TOKEN_\"");
+            cleanCsSource = RegexLafPublisherAssignment().Replace(cleanCsSource, "\"_YOUR_PUBLISHER_ID_\"");
+        }
+
```
haoliuu, line 328: Does this mean for the preview code in AI Dev Gallery, we are still using the existing pattern for calling LAF api?

weiyuanyue: Yes, this line of code will only be replaced when exporting. If only running in AI Dev Gallery, the execution will keep in prev code and still be 'LimitedAccessfeaturesHelper. xxxx'.

PR #476, opened 2025-10-22
Fix High Contrast Problem

wang563681252:
image

TO
image

### DIFF
--- a/AIDevGallery/Controls/ModelPicker/ModelOrApiPicker.xaml
+++ b/AIDevGallery/Controls/ModelPicker/ModelOrApiPicker.xaml
@@ -16,7 +16,7 @@
     mc:Ignorable="d">
     <UserControl.Resources>
         <ThemeShadow x:Name="DialogShadow" />
-        <SolidColorBrush x:Key="AppSmokeBrush" Color="#66000000" />
+        <SolidColorBrush x:Key="AppSmokeBrush" Color="#4D000000" />
     </UserControl.Resources>
     <Grid>
         <Grid


AIDevGallery/Controls/ModelPicker/ModelOrApiPicker.xaml
```diff
@@ -16,11 +16,13 @@
     mc:Ignorable="d">
     <UserControl.Resources>
         <ThemeShadow x:Name="DialogShadow" />
+        <SolidColorBrush x:Key="AppSmokeBrush" Color="#66000000" />
     </UserControl.Resources>
     <Grid>
         <Grid
             x:Name="ShadowGrid"
-            Background="{ThemeResource SmokeFillColorDefaultBrush}"
+            Background="{ThemeResource AppSmokeBrush}"
```
weiyuanyue, line 24: 🤔I am a bit concerned that using #66000000(semi transparent) could cause issues later. Windows High Contrast mode is designed to **maximize visibility** and HighContrast themes are **meant to make things super clear**, and a semi transparent color could make text harder to read. From the bug description, it seems that in this scenario, keeping the content behind the dialog visible in high contrast mode might have some intended significance🤔. If the change causes some elements to become unclear in High Contrast mode, it may still fail to meet accessibility standard [MAS 4.3.1 – No Disruption of Accessibility Features](https://aka.ms/MAS4.3.1)

wang563681252: fixed

PR #486, opened 2025-11-03
Fix markdown link click error

wang563681252:
ADO ID : 59122776

### DIFF
--- a/AIDevGallery/Pages/Models/ModelPage.xaml.cs
+++ b/AIDevGallery/Pages/Models/ModelPage.xaml.cs
@@ -17,7 +17,6 @@ using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using Windows.ApplicationModel.DataTransfer;
-using static System.Runtime.InteropServices.JavaScript.JSType;
 
 namespace AIDevGallery.Pages;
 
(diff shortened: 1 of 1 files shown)


AIDevGallery/Pages/Models/ModelPage.xaml.cs
```diff
@@ -302,14 +303,71 @@ private void CopyButton_Click(object sender, RoutedEventArgs e)
 
     private void MarkdownTextBlock_OnLinkClicked(object sender, CommunityToolkit.Labs.WinUI.MarkdownTextBlock.LinkClickedEventArgs e)
     {
-        string link = e.Url;
+        string? raw = e?.Url?.Trim();
 
-        ModelDetailsLinkClickedEvent.Log(link);
-        Process.Start(new ProcessStartInfo()
+        ModelDetailsLinkClickedEvent.Log(raw ?? "<null>");
```
haoliuu: Should we consider moving this log to a later point in the process if there is an error dialog or exception, so we can ensure the issue is logged correctly instead of logging "null"?

haoliuu (approved):

AIDevGallery/Pages/Models/ModelPage.xaml.cs
wang563681252: fixed

PR #494, opened 2025-11-06
Add AppContentSearch Samples

karkarl:
- Update WindowsAppSDK version to 2.0exp2
- AppContentSearch Samples: Semantic Search, Retrieval Augmented Generation, and Index Capability page.
- Global search with AppContentSearch
     - Currently disabled by default to avoid crash on launch if models aren't available.

### DIFF
--- a/AIDevGallery/Samples/WCRAPIs/AppIndexCapability.xaml
+++ b/AIDevGallery/Samples/WCRAPIs/AppIndexCapability.xaml
@@ -61,6 +61,18 @@
                     <TextBlock Grid.Column="1" x:Name="indexSemanticImageCapabilityResultText" Text="Unavailable"/>
                 </Grid>
             </StackPanel>
+            
+            <InfoBar x:Name="IndexCapabilitiesMessage"
+                     IsOpen="False"
+                     Severity="Warning"
+                     IsClosable="False"
+                     Title="Some index capabilities are unavailable."
+                     Padding="0">
+                <InfoBar.ActionButton>
+                    <HyperlinkButton Content="Windows AI Troubleshooting"
+                                     NavigateUri="https://learn.microsoft.com/en-us/windows/ai/apis/troubleshooting" />
+                </InfoBar.ActionButton>
+            </InfoBar>
         </StackPanel>
     </Grid>
 </samples:BaseSamplePage>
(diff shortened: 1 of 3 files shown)


karkarl:
> Could you please provide some usage instructions (e.g. device requirement) and steps to test about the new samples you added? I tried to install the artifacts on my device but looks like it doesn't work. Do we need to download something? 

Added a link to the troubleshooting page 

image

AIDevGallery/Samples/WCRAPIs/KnowledgeRetrieval.xaml
```diff
@@ -0,0 +1,237 @@
...
+                Foreground="{ThemeResource TextFillColorSecondaryBrush}" 
+                Text="Sample text data:"/>
+
+            <Grid>
+                <Button
+                    Margin="3,0,0,0"
+                    Click="AddTextDataButton_Click">
+                    <StackPanel Orientation="Horizontal">
+                        <FontIcon FontFamily="Segoe MDL2 Assets" Glyph="&#xE898;" FontSize="16" Margin="0,0,4,0"/>
+                        <TextBlock Text="Add Text Data"/>
```
haoliuu, line 88: One general feedback to the UX in RAG page: when the "Add Text Data" button is clicked, a new item is added after the last existing item. However, since there are already five items displayed horizontally, users cannot see the newly added item, and there is no way to scroll horizontally with the mouse to view it if they are not notified.
image

wang563681252: I think the sliders for these items are also a bit small.

wang563681252:
These UIs do not display completely in small windows. I suggest moving the 'Other samples' on the right to the bottom when the window is below a certain resolution.
image

AIDevGallery/Samples/WCRAPIs/KnowledgeRetrieval.xaml
karkarl: Great suggestion. These fixes can be addressed in a subsequent PR.

karkarl:
> These UIs do not display completely in small windows. I suggest moving the 'Other samples' on the right to the bottom when the window is below a certain resolution.

Great callout. I can address this in a subsequent PR.

haoliuu (approved):

PR #498, opened 2025-11-11
Video Super Resolution sample

huichen123:


AIDevGallery/Samples/WCRAPIs/VideoSuperRes.xaml.cs
```diff
@@ -0,0 +1,249 @@
...
+        CameraPreviewControl.PreviewFailed += CameraPreviewControl_PreviewFailed!;
+        await CameraPreviewControl.StartAsync(cameraHelper!);
+        CameraPreviewControl.CameraHelper.FrameArrived += CameraPreviewControl_FrameArrived!;
+        CameraPreviewControl.SizeChanged += CameraPreviewControl_SizeChanged;
+    }
+
+    private void CameraPreviewControl_PreviewFailed(object sender, PreviewFailedEventArgs e)
+    {
+        // Handle preview failure
+    }
```
haoliuu, line 111: Do we need to add the preview handle logic here in this placeholder comment? I encountered an error while testing locally and have shared the details with you in our chat.

AIDevGallery/Controls/SampleContainer.xaml.cs
```diff
@@ -70,6 +71,9 @@ public List<string> NugetPackageReferences
 
     private static readonly List<WeakReference<SampleContainer>> References = [];
 
+    [DllImport("onnxruntime.dll")]
+    private static extern IntPtr OrtGetApiBase();
+
```
haoliuu, line 76: Is this just for your local testing? If yes, would it make sense to remove it?

AIDevGallery/Samples/WCRAPIs/VideoSuperRes.xaml.cs
huichen123: your failure is probably due to the need to install EP. (are you on an internal build?) it won't be caught here anyway.

AIDevGallery/Controls/SampleContainer.xaml.cs
huichen123: this is a workaround to avoid a crash in onnxruntime.

AIDevGallery/Samples/WCRAPIs/VideoSuperRes.xaml.cs
```diff
@@ -0,0 +1,249 @@
...
+    {
+        try
+        {
+            await InitializeCameraPreviewControl();
+
+            var catalog = ExecutionProviderCatalog.GetDefault();
+            await catalog.EnsureAndRegisterCertifiedAsync();
+
+            var readyState = VideoScaler.GetReadyState();
+            if (readyState == AIFeatureReadyState.NotReady)
```
rodrigovaras, line 79: You should handle the readyState == NotSupportedOnCurrentSystem and provide a message

haoliuu (approved):

PR #511, opened 2025-11-17
Allow user to select flavor for Describe Your Change

ridizon-msft:
This change modifies the Windows AI Describe Your Change sample in Text Samples. In addition to letting a user provide a custom tone, from RewriteAsync, we also have a few prescribed flavors that the user should be able to select from: General, Casual, Concise and Formal.

The sample has been updated to allow a user to select any of the flavors above (using `RewriteAsync`) or to write a custom tone (using `RewriteCustomAsync`) which will have the similar behavior to the original sample.

- DescribeYourChange.xaml adds in radio buttons to select between the different flavors
- DescribeYourChange.xaml adds support for the original rewrite flavors from the `TextRewriteTone` enum and private functions for each of them when different tones are selected. When we do a rewrite, ensure that the previous radio button is still selected and grays out the Custom bar if the radio is not selected

This was tested by deploying locally via Visual Studio, using various kinds of custom tone and the preset flavors.

image

_Fig. 1: Screenshot showing the sample page selecting one of the prescribed tones, "Casual", when rewriting input text_

image

_Fig. 2: Screenshot showing the sample page with an example of how a user can use custom tone when rewriting input text_

### DIFF
--- a/AIDevGallery/Samples/WCRAPIs/DescribeYourChange.xaml.cs
+++ b/AIDevGallery/Samples/WCRAPIs/DescribeYourChange.xaml.cs
@@ -154,7 +154,7 @@ internal sealed partial class DescribeYourChange : BaseSamplePage
         return "General"; // Default radio fallback
     }
 
-    public async Task RewriteTextCustom(string prompt, string customTone)
+    public async Task RewriteTextCustom(string prompt)
     {
         if (_textRewriter == null)
         {
@@ -279,6 +279,8 @@ internal sealed partial class DescribeYourChange : BaseSamplePage
         // Disable the custom tone text box when Custom is deselected
         if (CustomToneTextBox != null)
         {
+            // Clear the text box to avoid data retention
+            CustomToneTextBox.Text = string.Empty;
             CustomToneTextBox.IsEnabled = false;
         }
     }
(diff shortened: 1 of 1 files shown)


AIDevGallery/Samples/WCRAPIs/DescribeYourChange.xaml.cs
```diff
@@ -203,16 +257,38 @@ public async Task RewriteTextCustom(string prompt, string customTone)
...
+    }
+
+    private void CustomRadioButton_Unchecked(object sender, RoutedEventArgs e)
+    {
+        // Disable the custom tone text box when Custom is deselected
+        if (CustomToneTextBox != null)
+        {
+            CustomToneTextBox.IsEnabled = false;
+        }
+    }
```
haoliuu, line 286: nit: please consider clearing CustomToneTextBox.Text when the radio is unchecked to avoid unintended data retention.

AIDevGallery/Samples/WCRAPIs/DescribeYourChange.xaml.cs
```diff
@@ -126,6 +127,33 @@ public bool IsProgressVisible
...
+        else if (CustomRadioButton.IsChecked == true)
+        {
+            // Empty custom text box will not generate a result
+            return string.IsNullOrWhiteSpace(CustomToneTextBox.Text) ? string.Empty : CustomToneTextBox.Text;
+        }
+
+        return "General"; // Default radio fallback
+    }
+
     public async Task RewriteTextCustom(string prompt, string customTone)
```
haoliuu: This customTone parameter is unused? Please consider to remove this parameter if it's not needed

ridizon-msft: Good call, `GetSelectedTone()` solves what we initially needed to pass here so I agree that this should be removed

haoliuu (approved):

Issue #523, opened 2025-12-03
[BUG]Incompatible locally imported models remain inaccessible and undeletable


weiyuanyue:
Currently, when users import a local ONNX model into AI Dev Gallery, the compatibility check is only performed after the model has been successfully imported. For models requiring DML/GPU acceleration on x64 devices, the compatibility logic enforces strict requirements:

1. A dedicated GPU must be present
2. Available VRAM must exceed model size by at least 2GB (for compatibility)

If these requirements are not met, the model is marked as [NotCompatible]. However, this creates a critical UX issue:

### Current Behavior

1. User successfully imports a local model file → ✅ Import succeeds
2. System evaluates compatibility post-import → ❌ Model marked as [NotCompatible](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
3. Model becomes invisible in the Model Management UI
4. User cannot use the model (expected)
5. User cannot delete the model (problematic)

### One of the following approaches should be implemented:

**Option 1 : Pre-import Validation**

- Perform compatibility checks before allowing model import
- Display clear warning messages if hardware requirements are not met
- Allow users to proceed with import only after acknowledging the incompatibility

**Option 2: Post-import Management**

- Allow incompatible models to remain visible in Model Management UI
- Display compatibility status clearly (e.g., with warning badges)
- Enable deletion of incompatible models

**Option 3: Hybrid Approach**

- Pre-import warning with user confirmation
- Post-import visibility with clear status indicators

### Impact

- User Confusion: Users don't understand why their successfully imported model is nowhere to be found
- Storage Waste: Incompatible models consume disk space but cannot be managed
- Poor UX: No feedback loop or recovery mechanism for users who inadvertently import incompatible models

PR #524, opened 2025-12-03
[Fix]Incorrect CPU Tagging for AITK NPU Models by Enhancing EP Detection

weiyuanyue:
AITK-converted NPU models were incorrectly tagged as CPU because the detection logic couldn't parse diverse execution provider configurations from `genai_config.json`.

## Root Cause
The existing implementation had several limitations:

- **Rigid schema parsing**: GenAIConfig.cs used fixed Dml and Cuda classes, unable to handle diverse execution providers (OpenVINO, VitisAI, QNN, etc.)
- **Incomplete provider scanning**: Only checked decoder-level session options, missing pipeline-level provider configurations
- **Missing NPU detection**: No runtime capability to verify if NPU hardware is actually available on the device

## Before/After

**Before**: NPU model shows CPU tag  

**After**: NPU model correctly tagged  

**Warning dialog** for unavailable EPs:  

## Solution

This PR implements a comprehensive fix with 4 key improvements:

1. Enhanced GenAI Config Parsing (GenAIConfig.cs)

    - Extended ProviderOptions class with flexible JSON extension data parsing to support various execution providers (DML, QNN, OpenVINO, VitisAI, etc.)
    - Added pipeline-level session options support, as AITK-converted models may specify execution providers at both decoder and pipeline levels
    - Implemented dynamic provider detection using `HasProvider()` and `GetProviderOptions()` methods to handle different config structures

2. Robust Hardware Detection (DeviceUtils.cs, new ExecutionProviderNames.cs)

    - Added `HasNPU()` method using ONNX Runtime's execution provider device detection
    - Implemented `GetEpDevices()` to query all available EPs on the device by ensuring certified execution providers are registered before detection
    - Introduced caching mechanism to avoid repeated EP registration overhead
    - Created centralized `ExecutionProviderNames` constants class to eliminate hardcoded EP strings and ensure consistency across the codebase

3. Intelligent Model Classification and Validation (UserAddedModelUtil.cs, ModelCompatibility.cs)

    - Completely rewrote `GetHardwareAcceleratorFromConfig()` to:
        - Check both decoder-level and pipeline-level provider options
        - Properly detect NPU configurations (QNN, OpenVINO with NPU device type, VitisAI)
        - Follow correct priority order: QNN > DML > NPU > GPU > CPU
    - Added `ValidateExecutionProviders()` to validate required EPs are available before model loading
    - Added NPU compatibility checks in `ModelCompatibility.cs` with proper error messaging
    - Implemented user-friendly warning dialog when loading models with unavailable execution providers

4. Code Quality Improvements

    - Standardized EP name references using `ExecutionProviderNames` constants throughout (`ScenarioPage.xaml.cs`, `WinMLHelpers.cs`)
    - Consolidate EP device retrieval to use DeviceUtils

## Testing

- AITK-converted NPU models display the correct NPU tag
- Models requiring unavailable execution providers show appropriate warnings
- The application gracefully handles various GenAI config structures
- Hardware detection is efficient and consistent

# Notes
**In the current workflow introduced by this PR, the early validation during model addition only checks for EP availability (e.g., attempting to load QNN EP on an Intel machine will trigger a warning dialog). It does not include hardware capability checks (such as verifying the size of dedicated video memory). Hardware capability validation occurs later during model loading, where incompatible models are hidden.
Future work will involve refining compatibility rules and improving the timing of these checks.**
Known related issue #523 (will not be fixed in this PR)

### DIFF
--- a/AIDevGallery/Utils/DeviceUtils.cs
+++ b/AIDevGallery/Utils/DeviceUtils.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.ML.OnnxRuntime;
 using System;
+using System.Linq;
 using Windows.Win32.Foundation;
 using Windows.Win32.Graphics.Dxgi;
 
@@ -114,13 +115,14 @@ internal static class DeviceUtils
     public static bool IsArm64() =>
         System.Runtime.InteropServices.RuntimeInformation.OSArchitecture == System.Runtime.InteropServices.Architecture.Arm64;
 
-    public static bool HasNPU()
+    public static bool HasNPU() => HasExecutionProvider(device =>
+        device.HardwareDevice.Type.ToString().Equals("NPU", StringComparison.OrdinalIgnoreCase));
+
+    private static bool HasExecutionProvider(Func<OrtEpDevice, bool> predicate)
     {
         try
         {
-            using SessionOptions options = new();
-            options.SetEpSelectionPolicy(ExecutionProviderDevicePolicy.PREFER_NPU);
-            return true;
+            return GetEpDevices().Any(predicate);
         }
         catch
         {


weiyuanyue:
@copilot

AIDevGallery/Models/GenAIConfig.cs
```diff
@@ -33,18 +39,53 @@ internal class GenAISessionOptions
...
+    [JsonExtensionData]
+    public Dictionary<string, JsonElement>? Stages { get; set; }
 }
 
-internal class Dml
+internal class PipelineStage
 {
+    [JsonPropertyName("session_options")]
+    public GenAISessionOptions? SessionOptions { get; set; }
 }
```
haoliuu, line 53: Does ``session_options`` in ``pipeline`` indicate the stage here?

AIDevGallery/Samples/SharedCode/WinMLHelpers.cs
```diff
@@ -12,32 +13,32 @@ namespace AIDevGallery.Samples.SharedCode;
...
-                case "OpenVINOExecutionProvider":
+                case ExecutionProviderNames.OpenVINO:
                     var device = devices.Where(d => d.HardwareDevice.Type.ToString().Equals(deviceType, StringComparison.Ordinal)).FirstOrDefault();
                     sessionOptions.AppendExecutionProvider(environment, [device], epOptions);
                     return true;
-                case "QNNExecutionProvider":
+                case ExecutionProviderNames.QNN:
                     // Configure performance mode for QNN EP
                     epOptions["htp_performance_mode"] = "high_performance";
                     break;
```
haoliuu, line 44: Should we handle ``AMD`` here?

AIDevGallery/Models/ModelCompatibility.cs
```diff
@@ -53,11 +53,27 @@ public static ModelCompatibility GetModelCompatibility(ModelDetails modelDetails
...
             compatibility = ModelCompatibilityState.Compatible;
         }
+        else if (modelDetails.HardwareAccelerators.Contains(HardwareAccelerator.QNN) && DeviceUtils.IsArm64())
+        {
+            compatibility = ModelCompatibilityState.Compatible;
+        }
+        else if (modelDetails.HardwareAccelerators.Contains(HardwareAccelerator.NPU))
+        {
+            // Check if any NPU is available using ONNX Runtime's EP detection
+            if (DeviceUtils.HasNPU())
```
haoliuu, line 67: Can we use WinML to support such case?

AIDevGallery/Utils/DeviceUtils.cs
```diff
@@ -107,8 +112,87 @@ public static ulong GetVram()
...
+    }
+
+    /// <summary>
+    /// Gets the list of available ONNX Runtime Execution Provider devices.
+    /// This method ensures that certified EPs (like OpenVINO, QNN, DML) are registered before querying,
+    /// as OrtEnv.GetEpDevices() only returns already-registered providers.
+    /// Results are cached to avoid repeated registration overhead.
+    /// </summary>
+    /// <returns>A read-only list of available ONNX Runtime Execution Provider devices.</returns>
+    public static System.Collections.Generic.IReadOnlyList<OrtEpDevice> GetEpDevices()
```
haoliuu, line 140: This naming could be a bit confusing compared to ``GetAvailableExecutionProviders``. People without context might not know which func to use.

AIDevGallery/Utils/DeviceUtils.cs
```diff
@@ -107,8 +112,87 @@ public static ulong GetVram()
...
+            var epDevices = GetEpDevices();
+            return epDevices
+                .Select(device => device.EpName)
+                .Distinct()
+                .ToList();
+        }
+        catch
+        {
+            return new System.Collections.Generic.List<string>();
+        }
```
haoliuu: Exception handling here may suppress any runtime errors. Should we log anything here?

AIDevGallery/Models/GenAIConfig.cs
weiyuanyue: image

Yes, `session_options` in the pipeline indicate the **specific execution stage**.

In AITK-converted models with decoder pipelines, each stage can have independent session options. 
In some model(like AMD & QNN):

- Both `context` and `iterator` stages explicitly specify `VitisAI` provider
- This ensures NPU acceleration for the computationally intensive transformer layers
- `embeddings` and `lm_head` don't need special provider config

My PR scans all pipeline stages because:

**1. Decoder-level `session_options` sometimes is empty `[]`**
2. The actual EP is only specified at pipeline stage level
3. Previous implementation only checked decoder level -> missed NPU detection

so the fix iterates through `decoder.Pipeline` to find provider config at any stage.

AIDevGallery/Utils/DeviceUtils.cs
weiyuanyue: Removed `GetAvailableExecutionProviders` method to reduce unnecessary abstraction

The GetEpDevices() method is retained because:
- It serves as an enhanced wrapper around OrtEnv.GetEpDevices() - Actually, I think OrtEnv.GetEpDevices() should be renamed, as it really means OrtEnv.GetRegisteredEpDevices(). Unregistered EPs (even though they exist) cannot be retrieved through this API :D
- It adds meaningful functionality (EP registration, caching, error handling)
- It maintains API consistency with the underlying ONNX Runtime naming

AIDevGallery/Utils/DeviceUtils.cs
weiyuanyue: thanks:D. Added a log here using the existing telemetry infra. Whether we use the success rate here as a metric can be planned later

AIDevGallery/Models/ModelCompatibility.cs
weiyuanyue: Thanks for the suggestion! The current code already uses WinML APIs where available:

WinML (`ExecutionProviderCatalog`) handles EP registration
ONNX Runtime (`OrtEnv.GetEpDevices()`) handles device enumeration

This combination is necessary because AFAIK WinML doesn't currently provide an API to query available devices. If you know of a WinML API that could replace `GetEpDevices()`, I'd be happy to use it!

AIDevGallery/Samples/SharedCode/WinMLHelpers.cs
weiyuanyue: No, we don't need to add a specific case for amd here. It doesn't require any special configuration options like DML, OpenVINO, or QNN do. The default branch already handles VitisAI correctly by calling AppendExecutionProvider with the standard parameters.

PR #527, opened 2025-12-08
[Feat]Add model configuration reset functionality to Settings page

weiyuanyue:
This PR adds a new "Reset model configuration" feature to the Settings page, allowing users to clear model-related configuration data without affecting downloaded model files.

### Background
The existing "Clear cache" functionality has limitations that can leave the application in an inconsistent state:

**Current Limitations:**
- "Clear cache" only deletes downloaded model files from disk
- It does NOT clear model configuration data stored in `state.json`
- Model preselection logic relies on this persistent configuration data

**Problem Scenario:**
When the app attempts to preselect a model based on usage history, it can crash if:
- **The model itself is corrupted, incompatible, or fails to load**
- The model is no longer available (external model moved/deleted, or cached model was cleared)

In these cases, users cannot access the model picker UI to change the selection, and clicking "Clear cache" doesn't help because it only manages physical files, not the configuration that determines preselection.

**Why This Matters:**
The app's preselection logic reads from `UsageHistoryV2` to automatically select the most recently used model. If this references a non-existent or problematic model, users need a way to clear this configuration independently of the model cache to recover.

### Changes
image

image

**UI (SettingsPage.xaml):**
- Added `SettingsCard` with "Reset" button in the "Model management" section
- Positioned before the "Model cache" expander

**Backend (SettingsPage.xaml.cs):**
- Implemented `ResetModelConfig_Click` to clear `UsageHistoryV2`, `ModelTypeToUserAddedModelsMapping`, and `MostRecentlyUsedItems` from `state.json`
- Added confirmation dialog before reset and completion dialog after success

Model configuration refers to preference data in `state.json` that controls how the app interacts with models -- separate from the model files themselves. This reset clears 3 types of data:

1. Usage History (`UsageHistoryV2`)
Records which models users previously selected and their preferred hardware accelerators. The model picker uses this to automatically preselect the most recently used model. After reset, it defaults to the first available model.

2. User-Added Models (`ModelTypeToUserAddedModelsMapping`)
Tracks external models manually added by users (like Ollama or custom ONNX models). After reset, these models disappear from selection lists and must be re-added.

3. Recent Items (`MostRecentlyUsedItems`)
Stores the last 6 scenarios and samples accessed, shown in the home page "Recent" section. After reset, this section becomes empty.

### Impact
**User Benefits:**
- Provides a recovery mechanism when model preselection fails due to stale references or problematic models
- Allows clearing configuration without losing downloaded models
- Enables recovery from crash loops where a bad model is automatically preselected

**Behavioral Changes:**
- Model selection UI falls back to default behavior (first available model) after configuration reset
- Clear separation: "Clear cache" (deletes files) vs. "Reset" (clears preferences)

**No Breaking Changes:**
- Downloaded model files remain untouched
- Other app settings are unaffected

### Testing
- Verify that `UsageHistoryV2`, `ModelTypeToUserAddedModelsMapping`, and `MostRecentlyUsedItems` are cleared
- Confirm `state.json` is updated correctly
- Test model preselection fallback behavior after reset
- Test recovery from crash scenarios with problematic models
- Verify downloaded models remain in cache

### DIFF
--- a/AIDevGallery/Pages/SettingsPage.xaml.cs
+++ b/AIDevGallery/Pages/SettingsPage.xaml.cs
@@ -133,15 +133,15 @@ internal sealed partial class SettingsPage : Page
         {
             // Clear usage history
             App.AppData.UsageHistoryV2?.Clear();
-            
+
             // Clear user-added model mappings
             App.AppData.ModelTypeToUserAddedModelsMapping?.Clear();
-            
+
             // Clear most recently used items
             App.AppData.MostRecentlyUsedItems.Clear();
-            
+
             await App.AppData.SaveAsync();
-            
+
             // Show confirmation
             ContentDialog confirmDialog = new()
             {


AIDevGallery/Pages/Scenarios/ScenarioPage.xaml.cs
```diff
@@ -246,32 +246,32 @@ private async void HandleModelSelectionChanged(List<ModelDetails?> selectedModel
...
+            SampleSelection.SelectedItem = visableSamples[0];
             SampleContainer.ShowFooter = true;
         }
         else
         {
             SampleContainer.ShowFooter = false;
-            LoadSample(viableSamples[0]);
+            LoadSample(visableSamples[0]);
         }
     }
```
haoliuu, line 276: Why was the name changed here? "visableSamples" seems like a spelling mistake, and the original "viableSamples" appears to be correct.

weiyuanyue: aha～got it. revert it/

haoliuu (approved):

PR #531, opened 2025-12-12
[Security] SHA-256 validation

haoliuu:
## Summary

This PR adds file integrity verification for model downloads from both Hugging Face and GitHub repositories using SHA-256 hashing.

### Core Verification Logic
- Added SHA-256 hash verification after model files are downloaded
- For **Hugging Face**: Extracts SHA-256 from the LFS `oid` field in the API response
- For **GitHub**: Detects Git LFS files and parses SHA-256 from LFS pointer files

image
image

### Model Download Enhancements
- Added `Verifying` status to the download progress states
- Added `VerificationFailed` status when hash mismatch is detected
- Users can choose to **keep** or **delete** the model if verification fails

### UI Updates
- Download progress list now shows verification status
- Added dialog for handling verification failures with clear options:
  - Delete model (recommended)
  - Keep anyway (user's choice)
  - Cancel (leave as-is)

### DIFF
--- a/AIDevGallery/Telemetry/Events/ModelIntegrityVerificationFailedEvent.cs
+++ b/AIDevGallery/Telemetry/Events/ModelIntegrityVerificationFailedEvent.cs
@@ -38,7 +38,7 @@ internal class ModelIntegrityVerificationFailedEvent : EventBase
     {
         TelemetryFactory.Get<ITelemetry>().LogError(
             "ModelIntegrityVerificationFailed_Event",
-            LogLevel.Error,
+            LogLevel.Info,
             new ModelIntegrityVerificationFailedEvent(modelUrl, fileName, verificationType, expectedValue, actualValue));
     }
 }


AIDevGallery/Controls/DownloadProgressList.xaml.cs
```diff
@@ -86,4 +87,60 @@ private void ClearHistory_Click(object sender, RoutedEventArgs e)
...
+                        },
+                        new TextBlock
+                        {
+                            Text = "Would you like to keep the model anyway or delete it?",
+                            TextWrapping = TextWrapping.Wrap
+                        }
+                    }
+                },
+                PrimaryButtonText = "Delete model",
+                SecondaryButtonText = "Keep anyway",
```
weiyuanyue: What is the security rationale for allowing users to keep potentially 
compromised models?

AIDevGallery/Controls/DownloadProgressList.xaml.cs
```diff
@@ -86,4 +87,60 @@ private void ClearHistory_Click(object sender, RoutedEventArgs e)
...
+                CloseButtonText = "Cancel",
+                DefaultButton = ContentDialogButton.Primary,
+                XamlRoot = this.XamlRoot
+            };
+
+            var result = await dialog.ShowAsync();
```
wang563681252: When you close the app under this logic, an intermediate state issue should occur.

AIDevGallery/Utils/ModelDownload.cs
```diff
@@ -187,6 +248,12 @@ private async Task<CachedModel> DownloadModel(string cacheDir, IProgress<float>?
                 var existingFileInfo = new FileInfo(existingFile);
                 if (existingFileInfo.Length == downloadableFile.Size)
                 {
+                    // Still need to verify existing files if they have a hash
+                    if (downloadableFile.ShouldVerifyIntegrity && downloadableFile.HasVerificationHash)
+                    {
+                        filesToVerify.Add((filePath, downloadableFile));
+                    }
+
                     continue;
```
weiyuanyue, line 220: If an existing cached file is corrupted or tampered, the download process will complete but the model will be unusable？

AIDevGallery/Utils/ModelDownload.cs
```diff
@@ -201,16 +268,67 @@ private async Task<CachedModel> DownloadModel(string cacheDir, IProgress<float>?
             var fileInfo = new FileInfo(filePath);
             if (fileInfo.Length != downloadableFile.Size)
             {
-                // file did not download properly, should retry
+                throw new IOException($"File size mismatch for {downloadableFile.Name}: expected {downloadableFile.Size}, got {fileInfo.Length}");
```
weiyuanyue: Where is this exception handled?

weiyuanyue: file size mismatch throws an IOException immediately, while SHA-256 verification failure allows user choice. 
Should we handle size mismatches similarly (with user choice) or should both be treated as hard failures?

AIDevGallery/Utils/ModelDownload.cs
```diff
@@ -201,16 +268,67 @@ private async Task<CachedModel> DownloadModel(string cacheDir, IProgress<float>?
...
+            }
+
+            if (FailedVerifications.Count > 0)
+            {
+                var failedFileNames = string.Join(", ", FailedVerifications.Select(f => f.FileName));
+                VerificationFailureMessage = $"Integrity verification failed for: {failedFileNames}";
+
+                // Cache the model info for potential KeepModelDespiteVerificationFailure call
+                var failedModelDirectory = url.GetLocalPath(cacheDir);
+                _pendingCachedModel = new CachedModel(Details, url.IsFile ? $"{failedModelDirectory}\\{filesToDownload.First().Name}" : failedModelDirectory, url.IsFile, modelSize);
```
weiyuanyue: When verification fails, we create `_pendingCachedModel` but don't add it to ModelCache until user explicitly chooses "Keep anyway". if the user clicks "Cancel" in the dialog, the files remain on disk but aren't tracked in the cache system and user can not manage them in AIDG Model Management, right?

AIDevGallery.Utils/ModelInformationHelper.cs
```diff
@@ -172,13 +214,26 @@ public static async Task<List<ModelFileDetails>> GetDownloadFilesFromHuggingFace
...
+                {
+                    sha256 = sha256.Substring(7);
+                }
+            }
+
+            return new ModelFileDetails()
             {
                 DownloadUrl = $"https://huggingface.co/{hfUrl.Organization}/{hfUrl.Repo}/resolve/{hfUrl.Ref}/{f.Path}",
-                Size = f.Size,
+                Size = f.Lfs?.Size ?? f.Size,
```
weiyuanyue: LFS size is always reliable and available? this will not break existing downloads where Lfs is null?

AIDevGallery.Utils/ModelFileDetails.cs
```diff
@@ -27,4 +29,24 @@ public class ModelFileDetails
...
+    /// Gets the expected SHA256 hash of the file.
+    /// For Hugging Face: from LFS oid field.
+    /// For GitHub: from LFS pointer file.
+    /// </summary>
+    public string? Sha256 { get; init; }
+
+    /// <summary>
+    /// Gets a value indicating whether this file should be verified for integrity (main model files like .onnx).
+    /// </summary>
+    public bool ShouldVerifyIntegrity => Name != null &&
```
weiyuanyue, line 43: Other critical files (tokenizers, config files, etc.) are not verified. Is this intentional?

AIDevGallery/Controls/DownloadProgressList.xaml.cs
haoliuu: Initially I didn't implement the "Keep" option, but I added it after considering that some users are developers who understand the risks. There are also legitimate scenarios where verification may fail (e.g., hash metadata out of sync on the source), and without this option, developers would have no way to use a model if verification consistently fails. The default action is "Delete" and keeping requires explicit user choice. Do you think we should not give developers any option to keep verification-failed files?

AIDevGallery/Utils/ModelDownload.cs
haoliuu: The cached file is still verified - it's added to filesToVerify and checked with SHA-256 later. If corrupted, DownloadStatus will be VerificationFailed and the user will be prompted.

AIDevGallery/Utils/ModelDownload.cs
haoliuu: Good catch. The original code had an empty if block here. I'm trying to address this issue for the legacy code. Size imcomplete sounds more like a hard failure to me since the files will not be used. Any thoughts?

AIDevGallery/Utils/ModelDownload.cs
haoliuu: Nice catch.  I've added cleanup logic to delete any verification-failed models that user didn't decide on.

AIDevGallery/Controls/DownloadProgressList.xaml.cs
haoliuu: Nice catch. I added cleanup logic to handle verification-failed models on app close.

AIDevGallery/Controls/DownloadProgressList.xaml.cs
weiyuanyue: makes sense. Since our target audience is developers, I agree we shouldn't block them completely.
To mitigate the risk, I suggest 2 improvements:
1.If user chooses to "keep anyway", its status should not simply change to **Completed** (which makes it look identical to a normal model). It will be better to display an “**Unverified**” label or a warning icon in the UI to remind users that this model carries potential risks.
2. The current warning msg is quite technical but doesn't clearly explain the security implications to users who may not fully understand what **match the expected hash** means. maybe we can consider like: "This could indicate file tampering or corruption. You will be using an unverified model that may behave unexpectedly"

AIDevGallery.Utils/ModelInformationHelper.cs
haoliuu: Good question. The ``??`` operator should handle this: if ``Lfs`` is null (non-LFS files), it falls back to f.Size, which is the original logic. This change here is for some of my local HF validation logic which has been removed. We can revert it back. Let me clean it up.

AIDevGallery.Utils/ModelFileDetails.cs
haoliuu: good catch. added

AIDevGallery/Controls/DownloadProgressList.xaml.cs
haoliuu: I refactored the validation workflow after discussion so that the model is no longer kept when validation fails. This should prevent the issue from happening again.

AIDevGallery/Controls/DownloadProgressList.xaml.cs
haoliuu: Discussed offline, we refactored the validation workflow so that the model is no longer kept when validation fails.

AIDevGallery/Utils/ModelDownload.cs
haoliuu: I did a refactor and treat the size and SHA validation in consistent manner.

AIDevGallery/Telemetry/Events/ModelIntegrityVerificationFailedEvent.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    public override void ReplaceSensitiveStrings(Func<string?, string?> replaceSensitiveStrings)
+    {
+    }
+
+    public static void Log(string modelUrl, string fileName, string verificationType, string expectedValue, string actualValue)
+    {
+        TelemetryFactory.Get<ITelemetry>().LogError(
+            "ModelIntegrityVerificationFailed_Event",
+            LogLevel.Info,
```
weiyuanyue: should be LogLevel.Error?

AIDevGallery/Utils/ModelDownload.cs
```diff
@@ -187,6 +211,12 @@ private async Task<CachedModel> DownloadModel(string cacheDir, IProgress<float>?
                 var existingFileInfo = new FileInfo(existingFile);
                 if (existingFileInfo.Length == downloadableFile.Size)
                 {
+                    // Still need to verify existing files if they have a hash
+                    if (downloadableFile.ShouldVerifyIntegrity && downloadableFile.HasVerificationHash)
+                    {
+                        filesToVerify.Add((filePath, downloadableFile));
```
weiyuanyue, line 217: if a file exists with the correct size, it is added to verification but the verification happens later with `filePath`, not `existingFile`. Is this correct? 
Should we verify existing files before deciding to skip the download?

haoliuu: The current design intentionally defers verification to after all downloads complete. This is for performance - SHA256 computation is I/O intensive, and we want to optimize the "resume download" scenario instead of doing SHA calculation for each file during the downloading process. filePath and existingFile are the same value here since we check ``f => f == filePath``, so using either one should be ok.

PR #538, opened 2025-12-17
[Test]Establish Comprehensive Testing Infra

weiyuanyue:
This PR establishes a comprehensive testing infrastructure for AI Dev Gallery by introducing three types of tests (unit, integration, and UI automation) with corresponding test frameworks and CI/CD integration.

##  Motivation

- **Quality Assurance**: Ensure code changes don't break existing functionality
- **Regression Prevention**: Catch UI and integration issues early in development
- **Maintainability**: Provide safety net for refactoring and feature additions
- **Documentation**: Tests serve as living documentation of expected behavior

### Project Restructure
- Renamed `AIDevGallery.UnitTests` → `AIDevGallery.Tests` to accommodate multiple test types
- Updated solution and project references accordingly

### Test Infrastructure
1. **FlaUITestBase**: High-level UI testing framework
   - Simplified API for common UI interactions
   - Automatic screenshot capture on test completion
   - Clean resource management

2. **NativeUIA3TestBase**: Low-level UI automation
   - Direct UIA3 COM interop for advanced scenarios
   - Detailed element inspection capabilities
   - Performance-critical test scenarios

3. **PerformanceCollector**: Test execution metrics
   - CPU, memory, and disk usage monitoring
   - Performance regression detection

#### Unit Tests
- **AppUtilsTests**: Validates utility functions (file size conversion, model metadata, etc.)
- **Auto-triggered on PRs** for fast feedback

#### Integration Tests
- **ProjectGeneratorTests**: End-to-end project generation workflow
- **Manual trigger** (dependency issues being resolved)

#### UI Automation Tests 🖱️
- **SmokeTests**: Quick sanity checks (app launch, window properties)
- **NavigationViewTests**: Navigation menu interactions
- **MainWindowTests**: Main window functionality
- **BasicInteractionTests**: Common user workflows
- **NativeUIA3SmokeTests**: Low-level automation validation
- **Manual trigger** to avoid blocking PR pipeline

### CI/CD Integration
Modified `.github/workflows/build.yml`:
-  Unit tests run automatically on every PR
-  Integration tests available via manual workflow dispatch
-  UI tests available via manual workflow dispatch

### Original Workflow(Before)

```mermaid
graph TD
    A[Trigger: Push/PR] --> B[Build]
    B --> C[Generate MSIX]
    C --> D[Upload Artifact]
    D --> E[Complete]
    
    style A fill:#e1f5ff
    style B fill:#fff4e1
    style E fill:#e8f5e9
```
### New Workflow(After)

```mermaid
graph TD
    A[Trigger Event] --> B{PR/Push?}
    
    B --> |Yes| C1[Run Unit Tests]
    C1 --> C2{Tests Pass?}
    C2 --> |Yes| C3[✓ Pass]
    C2 --> |No| C4[✗ Fail Fast]
    C4 --> C5[Generate Reports]
    
    B --> |Yes|D1[Build]
    D1 --> D2[Generate MSIX & Upload Artifact]
    D2 --> D3[✓ Complete]
    
    F[Download MSIX<br/>Skip Build]
    
    A --> E{Manual Trigger<br/>All Tests?}
    E -->|Yes| G[Build App & Tests<br/>Independent Mode]
    
    F --> H[Deploy MSIX]
    G --> H
    
    H --> I[Run All Tests<br/>Unit+Integration+UI]
    I --> J[Generate Reports]
    J --> K[✓ Complete]
    
    style A fill:#e1f5ff
    style F fill:#c8e6c9
    style G fill:#fff9c4
    style I fill:#e8f5e9
    style K fill:#c8e6c9
    style C4 fill:#ffcdd2
```
### Dependencies
- Added `FlaUI.UIA3` (v5.0.0) for UI automation framework

#  Known Limitations

- Integration tests currently blocked by dependency [resolution issues](https://github.com/microsoft/WindowsAppSDK/issues/5845)
- UI tests require manual trigger to avoid CI/CD overhead
- Performance baseline data still being collected

### DIFF
--- a/AIDevGallery.Tests/UITests/NavigationViewTests.cs
+++ b/AIDevGallery.Tests/UITests/NavigationViewTests.cs
@@ -7,7 +7,6 @@ using FlaUI.Core.Tools;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
 using System.Linq;
-using System.Threading;
 
 namespace AIDevGallery.Tests.UITests;
 


AIDevGallery.Tests/TestInfra/PerformanceCollector.cs
```diff
@@ -0,0 +1,283 @@
...
+/// 3. Memory tracking:
+/// <code>
+///   PerformanceCollector.TrackMemoryUsage(processId, "MemoryUsage_Startup");
+///   PerformanceCollector.TrackCurrentProcessMemory("MemoryUsage_Current");
+/// </code>
+/// </summary>
+public static class PerformanceCollector
+{
+    private static readonly JsonSerializerOptions JsonOptions = new() { WriteIndented = true };
+    private static readonly List<Measurement> _measurements = new();
```
wang563681252: Although the Add operation is protected by a lock, data may get mixed up when multiple test cases run simultaneously.

.github/workflows/test-report.yml
```diff
@@ -1,31 +1,163 @@
...
+  # TODO: Enable workflow_run trigger after validating the workflow works reliably
+  # This would allow automatic testing after successful CI builds
+  # workflow_run:
+  #   workflows: ["CI"]
+  #   types:
+  #     - completed
+
 jobs:
-  report:
+  regression-tests:
```
wang563681252, line 12: Maybe need add “needs: build”

AIDevGallery.Tests/IntegrationTests/ProjectGeneratorTests.cs
```diff
@@ -139,28 +139,35 @@ public static async Task Initialize(TestContext context)
         Directory.CreateDirectory(TmpPathProjectGenerator);
         Directory.CreateDirectory(TmpPathLogs);
 
-        TaskCompletionSource taskCompletionSource = new();
-
-        UITestMethodAttribute.DispatcherQueue?.TryEnqueue(() =>
+        if (UITestMethodAttribute.DispatcherQueue != null)
```
wang563681252, line 142: Maybe listView never have new ListView, please double check this code.

AIDevGallery.Tests/TestInfra/TestConfiguration.cs
```diff
@@ -0,0 +1,15 @@
...
+
+/// <summary>
+/// Central configuration for test infrastructure.
+/// </summary>
+public static class TestConfiguration
+{
+    /// <summary>
+    /// The MSIX package identity name GUID from Package.appxmanifest.
+    /// </summary>
+    public const string MsixPackageIdentityName = "e7af07c0-77d2-43e5-ab82-9cdb9daa11b3";
```
haoliuu: nit: would it be possible to read this GUID from the Package.appxmanifest file at runtime or from a shared configuration, instead of hardcoding? This ensures tests stay synchronized with the actual package identity.

AIDevGallery.Tests/TestInfra/FlaUITestBase.cs
```diff
@@ -0,0 +1,498 @@
...
+            // Use PowerShell to query installed packages
+            var startInfo = new ProcessStartInfo
+            {
+                FileName = "powershell.exe",
+                Arguments = $"-NoProfile -Command \"Get-AppxPackage | Where-Object {{$_.Name -like '*{TestConfiguration.MsixPackageIdentityName}*'}} | Select-Object -First 1 -ExpandProperty PackageFamilyName\"",
+                RedirectStandardOutput = true,
+                RedirectStandardError = true,
+                UseShellExecute = false,
+                CreateNoWindow = true
+            };
```
haoliuu, line 109: Looks like there are many duplicate methods here with ``NativeUIA3TestBase.cs``. Consider extracting shared functionality to a common utility class (e.g., TestHelper.cs) to avoid duplication.

AIDevGallery.Tests/UITests/NavigationViewTests.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        // Click each item
+        foreach (var item in navigationItems)
+        {
+            Console.WriteLine($"\nClicking navigation item: {item.Name}");
+
+            try
+            {
+                item.Click();
+                Thread.Sleep(1000);
```
haoliuu: Would it be possible to use explicit waits (e.g., ``Retry.WhileNull`` as used in ``MainWindowTests.cs``) instead of fixed Thread.Sleep? Fixed delays can cause flaky tests in CI environments where performance varies.

.github/workflows/build.yml
```diff
@@ -11,21 +11,22 @@ on:
...
       fail-fast: false
       matrix:
         include:
           - os: windows-2025
             dotnet-arch: 'x64'
             dotnet-configuration: 'Release'
-          - os: windows-11-arm
-            dotnet-arch: 'arm64'
-            dotnet-configuration: 'Release'
-    name: Build MSIX - win-${{ matrix.dotnet-arch }}
```
haoliuu, line 25: why removing the ``arm64`` support here? the ``build`` process still contains arm64.

AIDevGallery.Tests/TestInfra/PerformanceCollector.cs
haoliuu: Using static state for measurements might cause issues with parallel test execution. Also I'm wondering if the _measurements list might grows indefinitely during test runs.

.github/workflows/test-report.yml
```diff
@@ -1,31 +1,163 @@
...
   workflow_dispatch:
-  workflow_run:
-    workflows: ['CI'] # runs after CI workflow
-    types:
-      - completed
-permissions:
-  contents: read
-  actions: read
-  checks: write
+  # TODO: Enable workflow_run trigger after validating the workflow works reliably
```
haoliuu: For "validating the workflow works reliably", do you mean to merge this PR and test after that? Or we can uncomment this now

.github/scripts/deploy-msix.ps1
```diff
@@ -0,0 +1,120 @@
...
+    exit 1 
+}
+
+Write-Host "Using MSIX file: $($msixFile.FullName)"
+Write-Host ""
+
+# Cleanup previous installations
+Write-Host "Cleaning up previous installations..."
+Get-AppxPackage -Name "*AIDevGallery*" | Remove-AppxPackage -ErrorAction SilentlyContinue
+Get-AppxPackage -Name "*e7af07c0*" | Remove-AppxPackage -ErrorAction SilentlyContinue
```
haoliuu: I agree that Continue is correct behavior here. But we can consider logging a warning if the cleanup fails. Clean up failure might cause following installations fail. Logging helps with debugging.

AIDevGallery.Tests/UITests/MainWindowTests.cs
```diff
@@ -0,0 +1,301 @@
...
+    [TestCategory("UI")]
+    [Description("Verifies that the main window can be closed")]
+    public void MainWindowCanBeClosed()
+    {
+        // Arrange
+        Assert.IsNotNull(MainWindow, "Main window should be initialized");
+        Assert.IsTrue(MainWindow.IsAvailable, "Main window should be available before closing");
+
+        // Act
+        MainWindow.Close();
```
haoliuu, line 189: nit: This test closes the main window. Could it affect other tests by doing this?

.github/workflows/build.yml
```diff
@@ -36,37 +37,54 @@ jobs:
...
-        $manifestPath = "${{ github.workspace }}\AIDevGallery\Package.appxmanifest"
-        [xml]$manifest = get-content $manifestPath
-        $manifest.Package.Identity.Version = '${{ env.GitBuildVersionSimple }}.0'
-        $manifest.Save($manifestPath)
-    - name: Set LAF build props
+    - name: Cache NuGet packages
+      uses: actions/cache@v4
+      with:
+        path: ~/.nuget/packages
+        key: ${{ runner.os }}-${{ matrix.dotnet-arch }}-nuget-${{ hashFiles('**/packages.lock.json', '**/*.csproj') }}
```
haoliuu: Should we use ``Directory.Packages.props`` instead of ``**/*.csproj`` here?

.github/workflows/build.yml
weiyuanyue: It does not remove arm64 support. The **pr-unit-test** and **build** workflows are separate, the diff you see is a GitHub display issue. For pr-unit-test, running x64 only is sufficient because the tests are platform‑independent, and this helps reduce build time.

.github/workflows/test-report.yml
weiyuanyue: Thanks for the suggestion～ I don't think we need `needs: build` here, and here's why:
**The `needs` keyword only works within the same workflow file.** Since `build` is in [build.yml] and `regression-tests` is in [test-report.yml], we can't reference it across files --- GitHub would throw an error.
**For cross-workflow dependencies, we would use `workflow_run` instead.** We actually have this set up already (lines 7-10), just commented out for now. But here's the thing: we specifically don't want every build to trigger regression tests, right? The current setup gives us that control.

**What happens now:**

- build.yml runs on push/PRs automatically
- test-report.yml only runs when we manually trigger it
- When we do run it manually, it builds everything it needs from scratch

This separation is exactly what we want, regression tests on demand, not on every build.
If we added `needs` (even if it worked), it would force them to run together, which defeats the purpose of keeping them separate.

AIDevGallery.Tests/TestInfra/PerformanceCollector.cs
weiyuanyue: Good catch! Fixed by using `AsyncLocal`, each test context now has isolated measurements, preventing data mixing in parallel execution.

weiyuanyue: Both fixed! Used AsyncLocal for parallel test isolation, and Save() now auto-clears measurements to prevent memory growth.

.github/scripts/deploy-msix.ps1
weiyuanyue: Good catch! Added warning logs for cleanup failures while preserving the continue behavior. Also refactored build.yml to reuse this script.

.github/workflows/build.yml
weiyuanyue: Good catch! Fixed~

AIDevGallery.Tests/IntegrationTests/ProjectGeneratorTests.cs
weiyuanyue: Current situation: The Integration Test initialization has a known dependency issues, so `UITestMethodAttribute.DispatcherQueue` is null right now. I added this check to avoid compilation errors.

But it's actually correct design: This test was meant to support 2 modes:

- **With UI**: Shows a `ListView` with colored status updates and buttons to open build folders/logs (useful when running locally for a longtime)
- **Headless**: No UI, just runs tests (CI/CD mode)

When `DispatcherQueue` is null, `listView` stays null---that's expected. All uses of `listView` already have null-safety (`listView?.DispatcherQueue?.TryEnqueue`), so no crashes.

Started as a workaround for dependency issues, but it's the right pattern anyway. Should keep this check even after fixing the dependencies.

AIDevGallery.Tests/TestInfra/FlaUITestBase.cs
weiyuanyue: I will prefer keeping both classes separate for now because:

**Still evaluating**: **We're comparing both approaches to see which works better with WinUI3. Keeping them separate makes this easier.**
**Different type systems**: FlaUI uses different objects (Application, Window) than Native UIA3 (Process, IUIAutomationElement). Creating a shared helper would need messy generic wrappers that make things more complex, not less.
**Different cleanup**: FlaUI uses normal Dispose(), but Native UIA3 needs Marshal.ReleaseComObject(). Can't really unify these in a helper class.

we can remove one or refactor once we pick a winner or both are stable. sound reasonable?

AIDevGallery.Tests/TestInfra/TestConfiguration.cs
weiyuanyue: I originally kept it hardcoded coz **Keeping test infrastructure simple** is important for reliability. Reading XML at runtime adds dependencies and potential failure points.
This GUID is basically the project's "permanent ID" - it rarely changes. 

I agree we need flexibility when devs use different GUIDs for their local builds. So I've updated it to a hybrid approach.

This gives us the best of both worlds:

- Zero config for CI/CD and most devs (uses default)
- Flexibility for local dev (set env var to override)
- No XML parsing complexity
- Clear and explicit configuration method

```bash
# Default - just works
dotnet test

# Custom GUID for local dev
export TEST_PACKAGE_IDENTITY_NAME="your-custom-guid"
dotnet test
```

Does this approach work for you? I think it addresses both the sync concern and the local dev flexibility we need.

AIDevGallery.Tests/UITests/NavigationViewTests.cs
weiyuanyue: Thanks~ Fixed~

.github/workflows/test-report.yml
weiyuanyue: Removed it~

AIDevGallery.Tests/UITests/MainWindowTests.cs
weiyuanyue: `TestCleanup` will also try to close the window, but it's protected with try-catch so this test closing the window early won't cause issues. Each test gets its own Initialize/Cleanup cycle, so this won't affect other tests.

PR #544, opened 2026-01-05
Add Accessibility Test Pipeline

wang563681252:


AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,345 @@
...
+    [Description("Verifies accessibility compliance of multiple pages using Axe.Windows CLI")]
+    public void MultiPageAccessibilityAxeWindowsCliTest()
+    {
+        // Arrange
+        Assert.IsNotNull(MainWindow, "Main window should be initialized");
+
+        var appProcess = System.Diagnostics.Process.GetCurrentProcess();
+        var processId = appProcess.Id;
+
+        Console.WriteLine($"Testing app process ID: {processId}");
```
weiyuanyue, line 43: seems  `GetCurrentProcess()` returns the test runner's process ID rather than the AIDG application's process ID. Should the Axe.Windows CLI be scanning the application under test instead? How does this currently capture accessibility issues in AIDG UI?

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,345 @@
...
+// Licensed under the MIT License.
+
+using AIDevGallery.Tests.TestInfra;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace AIDevGallery.Tests.AccessibilityTests;
+
+[TestClass]
+public class AccessibilityTests : FlaUITestBase
```
weiyuanyue, line 15: This test class inherits from `FlaUITestBase` but doesn't seem to have `[TestInitialize]` or `[TestCleanup]` methods. How is the MainWindow being initialized before the test runs?

.github/workflows/build.yml
```diff
@@ -140,7 +140,143 @@ jobs:
         name: MSIX-${{ matrix.dotnet-arch }}
         path: ${{ github.workspace }}/AIDevGallery/AppPackages/*_${{ matrix.dotnet-arch }}_Test/AIDevGallery_*_${{ matrix.dotnet-arch }}.msix
 
-  accessibility-check:
+  Accessibility-Test:
```
weiyuanyue, line 156: I see there are now 2 accessibility-related jobs: `Accessibility-Test` and `Accessibility-Check`. They share many similar steps like MSIX download, verification, and Axe.Windows CLI setup. Could you help me understand the distinction between these 2 jobs? Is there a specific reason for keeping them separate rather than consolidating the logic?

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,345 @@
...
+            }
+
+            // Try to invoke the navigation item
+            try
+            {
+                if (navigationItem.Patterns.Invoke.IsSupported)
+                {
+                    navigationItem.Patterns.Invoke.Pattern.Invoke();
+                    Console.WriteLine($"Invoked navigation item: {pageName}");
+                    System.Threading.Thread.Sleep(2000); // Wait for page to load
```
weiyuanyue: could this timing be insufficient in slower CI env or with complex page loading? How confident are we that the page has fully loaded and is ready for accessibility scanning after this fixed wait?

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,345 @@
...
+                Assert.Fail($"Accessibility issues found on pages: {string.Join(", ", failedPages)}");
+            }
+            else
+            {
+                Console.WriteLine("✓ All pages passed accessibility checks");
+            }
+        }
+        catch (Exception ex)
+        {
+            Assert.Inconclusive($"Failed to run multi-page accessibility tests: {ex.Message}");
```
weiyuanyue: The catch block marks all exceptions as `Inconclusive`. Does this mean that actual a11y violations could be masked as inconclusive test results? How do we differentiate between infrastructure/environmental issues and genuine a11y failures that should fail the test?

.github/workflows/build.yml
```diff
@@ -140,7 +140,143 @@ jobs:
...
+    - name: Display Test Report
+      if: always()
+      uses: dorny/test-reporter@v2
+      with:
+        name: Accessibility Test Report - ${{ matrix.dotnet-arch }}
+        path: 'TestResults/*.trx'
+        reporter: dotnet-trx
+        fail-on-error: false
+
+  Accessibility-Check:
```
haoliuu: Add a comment here that this will be removed after a11y-test is completed

.github/workflows/build.yml
wang563681252: deleted Accessibility-Check

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: Deleted try catch.

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: In fact, it's enough to run it on the pipeline, since local machines are generally better than the pipeline.

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: Base class will run [TestInitialize] and [TestCleanup].

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: Changed to the correct code.

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,512 @@
...
+        var settingItem = MainWindow?.FindFirstDescendant(cf => cf.ByName("Settings"));
+
+        if (settingItem == null)
+        {
+            Console.WriteLine($"Navigation item '{pageName}' not found");
+            return false;
+        }
+
+        Thread.Sleep(200);
+        settingItem.Click();
```
weiyuanyue: Why always click Settings first?

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,512 @@
...
+                            if (!listItem.TryGetClickablePoint(out _))
+                            {
+                                Console.WriteLine($"Skipping sub-item {listItem.Name} as it is likely off-screen");
+                                continue;
+                            }
+
+                            Thread.Sleep(200);
+                            listItem.Click();
+                            Thread.Sleep(3000);
+                            listItem.Click();
```
weiyuanyue: Why click twice?

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,512 @@
...
+            Console.WriteLine($"Navigation item '{pageName}' not found");
+            return false;
+        }
+
+        // Find the navigation item by name
+        var settingItem = MainWindow?.FindFirstDescendant(cf => cf.ByName("Settings"));
+
+        if (settingItem == null)
+        {
+            Console.WriteLine($"Navigation item '{pageName}' not found");
```
weiyuanyue: Wrong message

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,512 @@
...
+            bool navigationSuccess = NavigateToPage(pageName);
+
+            if (!navigationSuccess)
+            {
+                Console.WriteLine($"Skipping accessibility scan for {pageName} due to navigation failure");
+                continue;
+            }
+
+            // Check if this page has list items to test
+            if (pageName == "Samples" || pageName == "Models" || pageName == "AI APIs")
```
weiyuanyue: The code checks for "Models" and "AI APIs" pages, but they're not in the pagesToTest array. Is this intentional, or should these pages be added to the test suite?

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: I should add a note that the purpose here is to prevent popup windows generated during testing from causing the target element click to fail.

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: I should add a note: the purpose here is to cancel the popup that appears when no model is selected or the model is not supported (because different pages implement popup styles differently, it is difficult to write a function to directly close the popup, so it is recommended to unify the popup format in the future).

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: You're right, these pages need to be added later. I will extract them into a list to store the pages that require a thorough check.

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: Fixed

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: Fixed

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,517 @@
...
+        MainWindow.Patterns.Window.Pattern.SetWindowVisualState(WindowVisualState.Maximized);
+
+        Assert.IsTrue(InitAxeWindows(), "Axe Init Failed");
+
+        // Get the actual process ID from the window, not the test runner process
+        var processId = MainWindow.Properties.ProcessId.Value;
+
+        Console.WriteLine($"Testing app process ID: {processId}");
+
+        // TODO: Next steps - nees add "Models" and "AI APIs" pages to the test list
```
weiyuanyue: TYPO: need to

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,517 @@
...
+            Console.WriteLine($"Navigation item '{pageName}' not found");
+            return false;
+        }
+
+        // Find the navigation item by name
+        var settingItem = MainWindow?.FindFirstDescendant(cf => cf.ByName("Settings"));
+
+        if (settingItem == null)
+        {
+            Console.WriteLine($"Setting item '{pageName}' not found");
```
weiyuanyue: Shoud be "Setting item not found"?

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,517 @@
...
+            return false;
+        }
+
+        // Prevents test-generated popups from blocking the target element click.
+        Thread.Sleep(200);
+        settingItem.Click();
+        Thread.Sleep(200);
```
weiyuanyue: I see many Thread.Sleep calls throughout the tests. While UI tests do require waiting, fixed sleep times might not be optimal.
Consider using the existing Retry.WhileTrue / Retry.WhileNull patterns, which continuously check conditions until satisfied or timeout.

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
weiyuanyue: The double-click pattern with hard waits might be brittle, what if the popup takes longer to appear/dismiss? Would a retry-based approach be more reliable?

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: Fixed

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: Fixed

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: Fixed

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: Retry-based methods are not more reliable; the most reliable approach is still to perform unified pop-up detection through a consistent pop-up format.

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,540 @@
...
+        }
+
+        // Prevents test-generated popups from blocking the target element click.
+        Retry.WhileTrue(
+            () =>
+            {
+                settingItem.Click();
+                return IsItemSelected(settingItem);
+            },
+            timeout: TimeSpan.FromSeconds(4));
```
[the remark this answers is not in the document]
wang563681252: fixed

.github/workflows/build.yml
```diff
@@ -236,120 +244,49 @@ jobs:
...
-      shell: powershell
+    - name: Cache NuGet packages
+      uses: actions/cache@v4
+      with:
+        path: ~/.nuget/packages
+        key: ${{ runner.os }}-${{ matrix.dotnet-arch }}-nuget-${{ hashFiles('**/packages.lock.json', 'Directory.Packages.props', '**/*.csproj') }}
+        restore-keys: |
+          ${{ runner.os }}-${{ matrix.dotnet-arch }}-nuget-
+    - name: Restore dependencies
+      run: dotnet restore AIDevGallery.sln -r win-${{ matrix.dotnet-arch }} /p:Configuration=${{ matrix.dotnet-configuration }} /p:Platform=${{ matrix.dotnet-arch }}
```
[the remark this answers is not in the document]
wang563681252: Fixed

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,540 @@
...
+
+        scanResults.Add(result);
+
+        if (!scanPassed)
+        {
+            failedPages.Add(pageName);
+        }
+
+        // Small delay between page tests
+        System.Threading.Thread.Sleep(1000);
```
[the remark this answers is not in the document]
wang563681252: There isn't a universal component right now, so for the time being, we're using sleep.

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,540 @@
...
+
+namespace AIDevGallery.Tests.AccessibilityTests;
+
+[TestClass]
+public class AccessibilityTests : FlaUITestBase
+{
+    /// <summary>
+    /// Gets or sets the test context which provides information about and functionality for the current test run.
+    /// </summary>
+    public TestContext TestContext { get; set; }
```
haoliuu: nit: the TestContext might be null since you used ```TestContext?.AddResultFile(file);``` in following code? How about  making it ``TestContext? TestContext { get; set; }`` or mark as ``required``?

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,540 @@
...
+        Retry.WhileTrue(
+            () =>
+            {
+                settingItem.Click();
+                return !IsItemSelected(settingItem);
+            },
+            timeout: TimeSpan.FromSeconds(4));
+        navigationItem.Click();
+        Console.WriteLine($"Clicked navigation item: {pageName}");
+        Thread.Sleep(2000); // Wait for page to load
```
haoliuu, line 247: Would ``Retry.WhileTrue`` works better than sleep here? It's already used in those UI Automation files we just added.

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
```diff
@@ -0,0 +1,540 @@
...
+        }
+
+        // Assert - All pages should pass
+        Console.WriteLine("\n=== Accessibility Test Summary ===");
+        foreach (var result in scanResults)
+        {
+            Console.WriteLine(result);
+        }
+
+        if (failedPages.Count > 0)
```
haoliuu, line 201: Should the failure be asserted here?

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: Fixed

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: Not really. I can't find any parameter to monitor the status of the current button.

AIDevGallery.Tests/AccessibilityTests/AppPageTest.cs
wang563681252: The conclusion of our discussion that day was that we wouldn't perform the assert here.

Issue #548, opened 2026-01-07
[BUG] App crashes when displaying model documentation with formatted links


weiyuanyue:
The application crashes with an `ArgumentException` when rendering markdown documentation that contains hyperlinks with embedded UI elements. 

<img width="1103" height="498" alt="Image" src="https://github.com/user-attachments/assets/864b0adc-40f5-498b-ad4f-65d102616584" />

### Steps to Reproduce

1. Launch AI Dev Gallery application
2. Navigate to **Models** page
3. Select **Generative** category
4. Click on **Stable Diffusion** model

**Result:** Application crashes immediately when the documentation loads.

### Expected Behavior

- Model documentation should render correctly with all hyperlinks functional
- Application should remain stable and responsive
- Users should be able to click on links to access external documentation
- No crashes or unexpected termination

### Actual Behavior

Application crashes with the following error:

```
Microsoft.UI.Xaml.dll!00007FFF0E21BFA4: 80070057 - E_INVALIDARG
Exception thrown: 'System.ArgumentException' in System.Private.CoreLib.dll
Value does not fall within the expected range.

Stack trace:
   at System.Runtime.InteropServices.Marshal.ThrowExceptionForHR(Int32 errorCode)
   at WinRT.DelegateExtensions.DynamicInvokeAbi(Delegate del, Object[] invoke_params)
   at ABI.System.Collections.Generic.IListMethods`2.AppendDynamic(IObjectReference obj, T value)
   at Microsoft.UI.Xaml.Documents.InlineCollection.Add(Inline item)
   at CommunityToolkit.Labs.WinUI.MarkdownTextBlock.TextElements.MyHyperlink.AddChild(IAddChild child)
```

- App Version: 0.5.4

### Additional Context

The crash occurs specifically when the markdown rendering engine attempts to add certain types of inline elements to hyperlink objects. The error code `0x80070057` (E_INVALIDARG) suggests an invalid argument is being passed to a WinUI component.

Debug output shows the crash happens when processing `InlineUIContainer` elements:

```
[MyHyperlink.AddChild] Attempting to add inline:
  - Inline type: InlineUIContainer
  - Current hyperlink Inlines count: 0
```

PR #549, opened 2026-01-07
[Fix]Crashes in model doc page: WinUI element hierarchy and UI thread reentrancy violations

### LINKED ISSUE #547: [BUG] App crashes with reentrancy error when clicking hyperlinks in model documentation
The application crashes with a fail-fast exception when users click on any hyperlink in the model documentation pages. This is caused by a WinUI 3 reentrancy violation when `Process.Start()` is called synchronously on the UI thread.

Image

## Steps to Reproduce

1. Launch AI Dev Gallery application
2. Navigate to **Models** page
3. Select **Generative** category
4. Click on **Stable Diffusion** model
5. Click on **any text link** in the documentation section

**Result:** Application immediately crashes with reentrancy error

## Expected Behavior

- Clicking a hyperlink should open the URL in the default browser
- Application should remain responsive and stable
- No crashes or error dialogs

## Actual Behavior

Application crashes with the following error:

```
WinRT originate error - 0x80004005: 'Reentrancy was detected in this XAML application. 
Use a debugger to locate the reentrant code and, if necessary, move that code to an 
asynchronous event handler. Press OK to exit the application.'

Microsoft.UI.Xaml.dll!00007FFF1C4D75A7: 8000FFFF - E_UNEXPECTED
Unhandled exception at 0x00007FFF1C59BB4D (Microsoft.ui.xaml.dll) in AIDevGallery.exe: 
0xC000027B: An application-internal exception has occurred
Unhandled exception at 0x00007FF856490830 (combase.dll) in AIDevGallery.exe: 
0xC0000602: A fail fast exception occurred
```

- **OS:** Windows 11
- **App Version: 0.5.4
- **Frequency:** 100% repro

weiyuanyue:
Fix #547 & #548 Both issues stem from fundamental misunderstandings of WinUI 3 framework constraints.

### 1. MyHyperlink Crash: Framework Element Hierarchy Violation

截屏2026-01-07 15 37 10

**The Core Problem:**  
WinUI 3's `Hyperlink` class enforces strict parent-child element relationships at the framework level. The application was attempting to add `InlineUIContainer` elements into hyperlinks, which WinUI explicitly prohibits. This is a fundamental architectural constraint of the WinUI framework - hyperlinks can only contain text-based inline elements, never UI containers.

**Why This Matters:**  
- WinUI validates element hierarchies at runtime and throws `ArgumentException` when violations occur
- The error code `0x80070057` (E_INVALIDARG) indicates the framework is rejecting an invalid parent-child relationship
- The original code had an empty `catch` block that silently swallowed these exceptions in some cases, masking the underlying problem
- When the exception wasn't caught properly, it caused immediate application termination

**Root Cause:**  
The markdown rendering pipeline was blindly attempting to add any inline element to hyperlinks without validating compatibility against WinUI's element hierarchy rules. This represents a gap between the markdown DOM model (which is flexible) and WinUI's strict XAML element tree constraints.

**Impact:**  
- 100% reproducible crash on all model documentation pages with formatted links

---

### 2. Reentrancy Violation: UI Thread Blocking in Event Handlers

截屏2026-01-07 15 37 37

**The Core Problem:**  
WinUI 3 has a strict threading model that prohibits blocking the UI thread, especially within XAML event handlers. The application was calling `Process.Start()` synchronously on the UI thread, which blocks until the external process is launched. This violates WinUI's reentrancy detection mechanism, which monitors for operations that could cause the UI message pump to become unresponsive.

**Why This Matters:**  
- WinUI 3's dispatcher queue has built-in reentrancy guards to prevent deadlocks and UI freezes
- When synchronous I/O or process operations block the UI thread, it can cause the message pump to stall
- The framework detects this condition and triggers a fail-fast exception (`0x80004005`, `0xC000027B`, `0xC0000602`) to prevent system-wide hangs
- This is a protective mechanism, not a bug - WinUI is enforcing architectural best practices

**Root Cause:**  
Event handlers were written assuming synchronous operations are safe on the UI thread. However, `Process.Start()` with `UseShellExecute = true` involves shell interaction and external process launching, which can take unpredictable amounts of time. This blocked the dispatcher queue, causing WinUI to detect reentrancy and terminate the application.

**Impact:**  
- Application crashes whenever users click hyperlinks in documentation

---

### Fix 1: Respect WinUI Element Hierarchy Constraints

Added pre-validation to filter out incompatible elements before attempting to add them to hyperlinks. The fix enforces WinUI's architectural rules in our code rather than relying on exception handling.

---

### Fix 2: Honor WinUI Threading Model

Moved all `Process.Start()` calls to background threads using `Task.Run()`, preventing UI thread blocking and reentrancy violations.

**Key Changes:**
- Wrapped 3 instances of `Process.Start()` in `Task.Run()`
- Used fire-and-forget pattern (discard `_`) since we don't need to await results
- Applied consistently across all external process launch scenarios

---

### Bonus: Simplified ScrollViewer Hierarchy

Removed redundant nested `ScrollViewer` in Models Page that created scrolling conflicts in the layout.

**Result:** Cleaner layout hierarchy with better scrolling behavior.

**Before**:
截屏2026-01-07 17 26 04
**After**:
截屏2026-01-07 17 24 05

### DIFF
--- a/AIDevGallery/Pages/Models/ModelPage.xaml.cs
+++ b/AIDevGallery/Pages/Models/ModelPage.xaml.cs
@@ -270,9 +270,10 @@ internal sealed partial class ModelPage : Page
                             UseShellExecute = true
                         });
                     }
-                    catch
+                    catch (Exception ex)
                     {
-                        // Silently fail if fallback also fails
+                        // Log the failure to open the fallback URL for diagnostics, but do not surface it to the user.
+                        Debug.WriteLine($"Failed to open AI Toolkit fallback URL: {ex}");
                     }
 
                     wasDeeplinkSuccesful = false;


AIDevGallery/Pages/Models/ModelPage.xaml
```diff
@@ -73,33 +73,28 @@
         </StackPanel>
 
 
-        <ScrollViewer
-            x:Name="RootScroller"
+        <Grid
+            x:Name="ContentGrid"
             Grid.Row="2"
```
haoliuu: not related to your change, but I feel this should be placed in ``Grid.Row="1"``?

AIDevGallery/Pages/Models/ModelPage.xaml.cs
```diff
@@ -252,19 +252,19 @@ private void ToolkitActionFlyoutItem_Click(object sender, RoutedEventArgs e)
...
             {
-                Process.Start(new ProcessStartInfo()
+                _ = Task.Run(() => Process.Start(new ProcessStartInfo()
                 {
                     FileName = "https://learn.microsoft.com/en-us/windows/ai/toolkit/",
                     UseShellExecute = true
-                });
+                }));
                 wasDeeplinkSuccesful = false;
             }
```
haoliuu: ``Task.Run`` is fire-and-forget so Exceptions thrown by ``Process.Start()`` inside the task will NOT be caught by the outer catch block. Consider moving the try-catch inside the Task? I know ``await`` may not satisfy your need here.

AIDevGallery/Pages/Models/ModelPage.xaml
weiyuanyue: Sharp eye!

AIDevGallery/Pages/Models/ModelPage.xaml.cs
weiyuanyue: Good point! Fixed by moving try-catch inside the Task.Run. Thank you! 👍

PR #559, opened 2026-01-12
[Fix]SINet model download URL

weiyuanyue:
## Description

This PR fixes the SINet model download issue by updating the model URL to use a specific commit hash instead of pointing to the main branch.

## Problem
image

The previous URL https://huggingface.co/qualcomm/SINet/blob/main/SINet.onnx was failing to download, causing users unable to download the SINet model.

## Solution

Updated the URL to point to a specific commit hash: https://huggingface.co/qualcomm/SINet/blob/c76eed22a7d493836189bf892e83ebf8c59152f6/SINet.onnx

This ensures:
- Stable and reliable download link
- Consistent model version
- No breaking changes from upstream updates

## Changes

- Modified \imagemodels.modelgroup.json\ to update SINet model URL

### DIFF
--- a/AIDevGallery/Samples/Definitions/Models/imagemodels.modelgroup.json
+++ b/AIDevGallery/Samples/Definitions/Models/imagemodels.modelgroup.json
@@ -295,7 +295,7 @@
           "SINet": {
             "Id": "b7cae321-478f-1177-9559-709df5dfe703",
             "Name": "SINet",
-            "Url": "https://huggingface.co/qualcomm/SINet/blob/c76eed22a7d493836189bf892e83ebf8c59152f6/SINet.onnx",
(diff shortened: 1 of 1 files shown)


.github/workflows/test-report.yml
```diff
@@ -88,9 +88,22 @@ jobs:
         if (-not [string]::IsNullOrWhiteSpace("${{ secrets.LAF_PUBLISHER_ID }}")) {
           echo "LAF_PUBLISHER_ID=${{ secrets.LAF_PUBLISHER_ID }}" >> $env:GITHUB_ENV
         }
+    - name: Create ephemeral nuget config from template
```
haoliuu, line 91: Do we need this for URL fix?

haoliuu (approved):

